PR #2, opened 2021-02-10
Adds API for creating Ray clusters

sonnysideup:
along with other scaffolded files and types (i.e. controller)


ddl-kevin (commented):
LGTM

api/v1alpha1/raycluster_types.go
```diff
@@ -0,0 +1,100 @@
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RayClusterSpec defines the desired state of a RayCluster resource.
+type RayClusterSpec struct {
```
sonnysideup, line 9: @ddl-kevin please review this API spec and let me know your thoughts. thanks.

ddl-kevin (commented):
A couple of notes / questions. I can't think of anything else I would include right now, but I'm sure we will find some things.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	// +kubebuilder:default={repository: "ray-project/ray-ml", tag: "1.2.0-cpu"}
+	Image OCIImageDefinition `json:"image"`
+
+	// WorkerReplicaCount configures the total number of workers in the cluster.
+	// +kubebuilder:default=2
+	WorkerReplicaCount int `json:"workerReplicaCount"`
+
+	// HeadPort is the port of the head ray process.
+	// +kubebuilder:default=6379
+	HeadPort int `json:"port"`
```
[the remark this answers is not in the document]
sonnysideup: My guess is that we will not override this value but at least it's there for users who want to modify the default behavior.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	RedisShardPorts []int `json:"redisShardPorts"`
+
+	// ObjectManagerPort is the raylet port for the object manager.
+	// +kubebuilder:default=2384
+	ObjectManagerPort int `json:"objectManagerPort"`
+
+	// NodeManagerPort is the raylet port for the node manager.
+	// +kubebuilder:default=2385
+	NodeManagerPort int `json:"nodeManagerPort"`
+
```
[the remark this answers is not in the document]
sonnysideup: I noticed those flags mentioned [here](https://docs.ray.io/en/latest/configure.html#all-nodes) but didn't feel the need to expose them yet. 🤷‍♂️

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	// +kubebuilder:default={6380,6381}
+	RedisShardPorts []int `json:"redisShardPorts"`
+
+	// ObjectManagerPort is the raylet port for the object manager.
+	// +kubebuilder:default=2384
+	ObjectManagerPort int `json:"objectManagerPort"`
+
+	// NodeManagerPort is the raylet port for the node manager.
+	// +kubebuilder:default=2385
+	NodeManagerPort int `json:"nodeManagerPort"`
```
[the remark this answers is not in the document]
sonnysideup: They can either be explicitly specified by the consumer or they can use the default I've assigned arbitrarily. Whenever someone creates a CR w/o these values specified, these default values will be populated before it's persisted within Kubernetes. This means an inspection of the CR after creation will reveal what port values are being used.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	Image OCIImageDefinition `json:"image"`
+
+	// WorkerReplicaCount configures the total number of workers in the cluster.
+	// +kubebuilder:default=2
+	WorkerReplicaCount int `json:"workerReplicaCount"`
+
+	// HeadPort is the port of the head ray process.
+	// +kubebuilder:default=6379
+	HeadPort int `json:"port"`
+
```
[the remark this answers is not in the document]
sonnysideup: Yup. I'm not sure this is used for external comms though, do you know? If not, then we probably do not need to expose an external configuration. I honestly don't know enough here to have an opinion yet and was planning on adding this field if/when needed.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	// +kubebuilder:default=8265
+	DashboardPort int `json:"dashboardPort"`
+
+	// EnableNetworkPolicy will create a network policy that restricts ingress to the cluster.
+	// +kubebuilder:default=true
+	EnableNetworkPolicy bool `json:"enableNetworkPolicy"`
+
+	// EnablePodSecurityPolicy will create a pod security policy that enforces secure execution of the ray process.
+	// +kubebuilder:default=true
+	EnablePodSecurityPolicy bool `json:"enablePodSecurityPolicy"`
```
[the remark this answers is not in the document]
sonnysideup: We will create a new service account (cheap operation) for each new cluster and attach permissions using that isolated resource. The idea here is that every cluster should be self-contained and not rely on the existence of external resources in order to function correctly.

We _could_ use external service accounts but that feels like a 2nd-step, and this does not fit in with the general approach of the platform team.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+
+	// Labels additionally applied to cluster nodes.
+	Labels map[string]string `json:"labels"`
+
+	// Annotations applied to cluster nodes.
+	Annotations map[string]string `json:"annotations"`
+
+	// Resources requested and limits applied to cluster nodes.
```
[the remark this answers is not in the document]
sonnysideup: Potentially. I haven't tested or benchmarked running different workloads through a cluster. If the resource demands of the different nodes are different, then I will definitely split this into two, separate fields.

api/v1alpha1/raycluster_types.go
```diff
@@ -1,31 +1,83 @@
...
+	Resources corev1.ResourceRequirements `json:"resources"`
+
+	// Tolerations applied to cluster nodes.
+	Tolerations []corev1.Toleration `json:"tolerations"`
 
-	// Foo is an example field of RayCluster. Edit RayCluster_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// InitContainers added to cluster node pods.
+	InitContainers []corev1.Container `json:"initContainers"`
 }
```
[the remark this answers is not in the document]
sonnysideup: I'm already planning on adding an `emptyDir` volume mounted to `"/dev/shm"` without it having to be specified since that's what is recommended by the Ray peeps.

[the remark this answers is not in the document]
sonnysideup: Say more. Are you talking about the ability to add volume mounts or something else?

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
sonnysideup: I don't understand the question, would you please elaborate.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: I have absolutely no idea what this does tbh.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: Yep, I guess my point is that this seems to be changing the default behavior of the application. At the very least we probably need to track this for documentation.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: Yeah, the default range is pretty large so I think we will need to dig into this.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: I think thats fine, it sounded like we would create a PSP for the cluster. I'm not sure if that is the correct approach.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: FWIW, the experimental Ray operator is very flexible here, including different pools of workers with different resources. I dont think that is necessary here but FYI.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: Yes, I think we will almost always want to mount some volume for local scratch space (like Spark does). We will probably want some way to configure the size.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
ddl-kevin: Ray will call frameworks which communicate out of band with Ray. They will introduce other ports that may need to be opened. I think, compared to Spark, this may be especially important for Ray if we plan to lock down network communication.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
sonnysideup: I'll figure it out and report back whether or not this is a concern.

api/v1alpha1/raycluster_types.go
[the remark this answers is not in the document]
sonnysideup: It just sounds like the network policies have to accommodate for those extra ports. I look to Torch as an example and see what that presents.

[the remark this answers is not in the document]
sonnysideup: This seems more relevant when running outside a container honestly. Within a container, the workers can allocate any number of ports they want without concern, and networking between all of the cluster nodes should not be an issue.

I'll revisit this as soon as I stand up a cluster and start poking around inside the nodes.

api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,31 @@
...
+	// Repository where the container image is stored.
+	Repository string `json:"repository"`
+
+	// Tag points to a specific container image variant.
+	Tag string `json:"tag"`
+
+	// PullPolicy used to fetch container image.
+	// +kubebuilder:validation:Optional
+	PullPolicy corev1.PullPolicy `json:"pullPolicy"`
+}
```
steved, line 31: Does this need a pull secret?

api/v1alpha1/raycluster_types.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// RayClusterSpec defines the desired state of a RayCluster resource.
+type RayClusterSpec struct {
+	// Image used to launch head and worker nodes.
+	// +kubebuilder:default={repository: "rayproject/ray", tag: "1.2.0-cpu"}
+	// +kubebuilder:validation:Optional
+	Image *OCIImageDefinition `json:"image"`
+
+	// Autoscaling blah blah
```
steved, line 15: update before final merge?

pkg/resources/ray/horizontalpodautoscaler.go
```diff
@@ -0,0 +1,57 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+
+	dcv1alpha1 "github.com/dominodatalab/distributed-compute-operator/api/v1alpha1"
+)
+
+const (
+	scaleTargetKind            = "Deployment"
+	stabilizationWindowSeconds = 60
```
steved, line 15: This window time feels like something that might end up being configurable because of cluster constraints

pkg/resources/ray/networkpolicy.go
```diff
@@ -0,0 +1,43 @@
...
+			Name:      fmt.Sprintf("%s-cluster", rc.Name),
+			Namespace: rc.Namespace,
+			Labels:    MetadataLabels(rc),
+			Annotations: map[string]string{
+				resources.DescriptionAnnotationKey: "Allows all ingress traffic between cluster nodes",
+			},
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: labelSelector,
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
```
steved, line 29: Is there no external ingress allowed at all (like workspaces or something)?

steved (commented):
Nothing controversial 👍

api/v1alpha1/raycluster_types.go
sonnysideup: ha. missed that, thanks.

pkg/resources/ray/horizontalpodautoscaler.go
sonnysideup: that's easy to make configureable. I was just starting with what I don't know but that seems like a reasonable field to expose and we can use a default value of 60 when not specified.

pkg/resources/ray/networkpolicy.go
sonnysideup: I just overlooked this. I need to add one more netpol to account for external access via a label (e.g. `ray-client: "true"`).

api/v1alpha1/common_types.go
sonnysideup: The only thing worth note here is that I created the pull secret field directly on the spec as a slice of ref objects. This way N number of pull secrets can be added to pull both the ray image and any additional init container images if/when they live in different private registries.

Makefile
```diff
@@ -28,10 +28,14 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
...
 
+# Set the shell used to bash for better error handling.
+SHELL = /bin/bash
+.SHELLFLAGS = -e -o pipefail -c
+
 # Image URL to use all building/pushing image targets
 IMG ?= controller:latest
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+# Produce CRDs that work with Kubernetes 1.16+ and support defaulting, version conversion, and field pruning
```
ddl-kevin, line 37: Are we planning to require 1.16 at a minimum?

api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,37 @@
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+// Autoscaling configuration for scalable workloads.
+type Autoscaling struct {
```
ddl-kevin, line 6: Are we planning to always deploy HPA or is this type intended to abstract whatever scaling is used?

ddl-kevin (commented):
This is looking good. It seems like some things are still in flux so I just added some questions.

Makefile
sonnysideup: I would say "yes" because of the following reasons:

- the current version is 1.20 and providers like EKS and the like only maintain support for 3-4 older versions
- 1.16 is stable and has been out for quite a while so programming to this version seems like a reasonable middle ground as opposed to supporting something like 1.12
- 1.16 gives use really good support for CRD versioning.

api/v1alpha1/common_types.go
sonnysideup: HPA is the standard way to scale Pods in K8s. It supports standard metrics like cpu/metrics as well as policies that support custom metrics if we ever want to invest in that.

ddl-kevin: It sounds like the answer is yes, then. Would it make sense to just use the HPAv2 Spec type here instead of defining our own?

Makefile
ddl-kevin: I suspect this will cause issues for a number of customers. I'm aware of some that have only just updated to 1.15.

pkg/resources/ray/networkpolicy.go
sonnysideup: I added the field `NetworkPolicyClientLabels` to the spec so that users can override the default metadata (e.g. `ray-client: "true`).

Makefile
sonnysideup: Platform has used new features to push clients to upgrade Kubernetes (for their benefit) in the past; it often works. I'm going to program with the expectation that we're running on 1.16+ and we'll figure out some graceful degradation path if and when.

api/v1alpha1/common_types.go
sonnysideup: We have the need for both short and long-lived cluster executions. In the former case, HPA will most likely not be enabled, whereas the latter seems more likely to want HPA.

I think reworking this to leverage `minReplicas` might make more sense but I don't feel inclined to expose the entire resource. The idea behind this sentiment is, if we understand how ray clusters work well, then we should be able to abstract some scaling details and expose only the fields that are relevant to different workloads.

sonnysideup:
@steved This is about 99% complete; I just need to add some more logging to the controller. I'd appreciate it if you would give this another look.

Makefile
```diff
@@ -43,78 +17,96 @@ else
...
+GOIMPORTS = $(shell pwd)/bin/goimports
+goimports: ## Download goimports locally if necessary.
+	$(call go-get-tool,$(GOIMPORTS),golang.org/x/tools/cmd/goimports)
+
+GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
+golangci-lint: ## Download golangci-lint locally if necessary.
+	@[ -f $(GOLANGCI_LINT) ] || { \
+		set -e ;\
+		echo "Installing golangci-lint" ;\
+		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(PROJECT_DIR)/bin v1.37.1 ;\
```
steved, line 107: extract the version still maybe?

api/v1alpha1/raycluster_webhook.go
```diff
@@ -0,0 +1,310 @@
...
+		r.Spec.Image = defaultImage
+	} else {
+		if r.Spec.Image.Repository == "" {
+			log.Info("setting default image repository", "value", defaultImage.Repository)
+			r.Spec.Image.Repository = defaultImage.Repository
+		}
+		if r.Spec.Image.Tag == "" {
+			log.Info("setting default image tag", "value", defaultImage.Tag)
+			r.Spec.Image.Tag = defaultImage.Tag
+		}
```
steved, line 115: This seems a bit odd, if I define the Image and want the repo or tag empty shouldn't it leave it be?

api/v1alpha1/raycluster_webhook.go
```diff
@@ -0,0 +1,310 @@
...
+func (r *RayCluster) validateWorkerReplicas() *field.Error {
+	replicas := r.Spec.Worker.Replicas
+	if replicas == nil || *replicas >= 0 {
+		return nil
+	}
+
+	return field.Invalid(
+		field.NewPath("spec").Child("worker").Child("replicas"),
+		replicas,
+		"should be greater than or equal to 0",
```
steved, line 182: Is this not potentially a desirable feature? E.g. could we support scale-down when not in use?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,424 @@
...
+	return nil
+}
+
+// reconcileAutoscaler optionally creates a horizontal pod autoscaler that
+// targets Ray worker pods.
+func (r *RayClusterReconciler) reconcileAutoscaler(ctx context.Context, rc *dcv1alpha1.RayCluster) error {
+	if rc.Spec.Autoscaling == nil {
+		hpa := &autoscalingv2beta2.HorizontalPodAutoscaler{
+			ObjectMeta: ray.HorizontalPodAutoscalerObjectMeta(rc),
+		}
```
steved, line 209: Similar to the other reconcile methods, why can't you pass in the real `hpa` object from L213?

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,424 @@
...
+	log.V(1).Info("applying patch to object", "object", controlled, "patch", string(patchResult.Patch))
+	if err = PatchAnnotator.SetLastAppliedAnnotation(controlled); err != nil {
+		return err
+	}
+
+	controlled.SetResourceVersion(found.GetResourceVersion())
+	if modified, ok := controlled.(*corev1.Service); ok {
+		current := found.(*corev1.Service)
+		modified.Spec.ClusterIP = current.Spec.ClusterIP
+	}
```
steved, line 292: this ... seems a little odd and scary (i.e. feels like we should either have immutable services or somehow copy all the stateful-ish fields)

controllers/raycluster_controller.go
```diff
@@ -0,0 +1,424 @@
...
+func (r *RayClusterReconciler) getLogger(ctx context.Context) logr.Logger {
+	if ctx == nil {
+		return r.Log
+	}
+	if ctxLogger, ok := ctx.Value(loggerKey).(logr.Logger); ok {
+		return ctxLogger
+	}
+
+	return r.Log
+}
```
steved, line 424: Is there no `ContextWithLogger` type we can leverage?

deploy/helm/distributed-compute-operator/templates/hooks.yaml
```diff
@@ -0,0 +1,144 @@
...
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade,post-delete
+    "helm.sh/hook-weight": "-3"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: dominodatalab:operator:distributed-compute-operator
```
steved, line 17: ideally, these global names have some sort of release/ns identifier since they can conflict (though maybe that's just not a thing given this operator needs other global resources)

deploy/helm/distributed-compute-operator/templates/hooks.yaml
```diff
@@ -0,0 +1,144 @@
...
+      {{- end }}
+      serviceAccountName: {{ include "dco.serviceAccountName" . }}-hook
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      restartPolicy: Never
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
steved, line 84: def need to use the image tag helper before we release it for reals

deploy/helm/distributed-compute-operator/templates/webhook-service.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "dco.webhook.service" . }}
+  labels:
+    {{- include "dco.labels" . | nindent 4 }}
+spec:
+  ports:
+    - port: 443
+      targetPort: 9443
```
steved, line 10: should we use named target ports? are named target ports passe?

deploy/helm/distributed-compute-operator/values.yaml
```diff
@@ -0,0 +1,79 @@
...
+  logEncoder: ""
+  # Logger level can be either 'debug', 'info', 'error', or any integer value >= 0 which corresponds to custom debug levels of increasing verbosity
+  logLevel: ""
+  # Logger level at which stacktraces are printed can either 'info', 'error', or 'panic'
+  logStacktraceLevel: panic
+  # Development mode enables debug logging, console output and stacktraces suitable for troubleshooting
+  logDevelopmentMode: false
+
+image:
+  repository: ghcr.io/dominodatalab/distributed-compute-operator
```
steved, line 31: ideally, still would like the registry separated from repo even for external releases

pkg/crd/crd.go
```diff
@@ -0,0 +1,132 @@
...
+// The latest available version of the CRD will be used to perform this operation.
+func Apply(ctx context.Context) error {
+	apply := func(client apixv1client.CustomResourceDefinitionInterface, crd *apixv1.CustomResourceDefinition) error {
+		found, err := client.Get(ctx, crd.Name, metav1.GetOptions{})
+
+		switch {
+		case apierrors.IsNotFound(err):
+			log.Info("Creating CRD", "Name", crd.Name)
+			_, err = client.Create(ctx, crd, metav1.CreateOptions{})
+		case err == nil:
```
steved, line 31: You use a different pattern elsewhere:
```
if apiErrors.IsNotFound(err) { ... }
if err != nil { ... }

...
```

Not sure which is best, but a single style would be better I think.

pkg/manager/manager.go
```diff
@@ -0,0 +1,84 @@
...
+	if err = (&controllers.RayClusterReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("RayCluster"),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "RayCluster")
+		return err
+	}
+
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
steved, line 53: Why is this an env variable rather than a CLI arg?

pkg/manager/manager.go
```diff
@@ -0,0 +1,84 @@
...
+		if err = (&dcv1alpha1.RayCluster{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "RayCluster")
+			return err
+		}
+	}
+	//+kubebuilder:scaffold:builder
+
+	if err := mgr.AddHealthzCheck("health", healthz.Ping); err != nil {
```
steved, line 61: TIL, cool

pkg/resources/ray/deployment.go
```diff
@@ -0,0 +1,252 @@
...
+	serviceAccountName := InstanceObjectName(rc.Name, ComponentNone)
+	if rc.Spec.ServiceAccountName != "" {
+		serviceAccountName = rc.Spec.ServiceAccountName
+	}
+
+	deploy := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Deployment",
+			APIVersion: "apps/v1",
+		},
```
steved, line 95: This isn't interpreted from the type of the ... type?

pkg/resources/ray/ray.go
```diff
@@ -0,0 +1,56 @@
...
+	ApplicationName = "ray"
+)
+
+// InstanceObjectName returns the name that will be used to create most owned cluster resources.
+func InstanceObjectName(instance string, comp Component) string {
+	if comp == ComponentNone {
+		return fmt.Sprintf("%s-%s", instance, ApplicationName)
+	}
+
+	return fmt.Sprintf("%s-%s-%s", instance, ApplicationName, comp)
```
steved, line 30: Do issues like a name being too long bubble up to the CRD or something?

steved (approved):
Left some comments, but I don't think any are a blocking issue.

pkg/manager/manager.go
sonnysideup: I know. This is pretty badass.

Makefile
sonnysideup: I had that before but decided to follow the pattern established by the rest of the installer targets. It's only declared and used in a single place so my thinking is _if you want to update a tool version, then go the tools section and update it._

I guess I'd prefer to leave it as-is until I actual need to reuse or output the version(s).

steved: Cool, that makes sense. As long as it was deliberate.

api/v1alpha1/raycluster_webhook.go
sonnysideup: ### Repo

```go
type OCIImageDefinition struct {
	// Registry where the container image is hosted.
	Registry string `json:"registry,omitempty"`

	// Repository where the container image is stored.
	Repository string `json:"repository,omitempty"`

	// Tag points to a specific container image variant.
	Tag string `json:"tag,omitempty"`

	// PullPolicy used to fetch container image.
	PullPolicy corev1.PullPolicy `json:"pullPolicy,omitempty"`
}
```

To my knowledge, it's fine if a `Registry` is empty, but a`Repository` cannot be. Are you thinking of the former?

### Tag

I didn't think of it before but I guess it's valid for `Tag` to be blank. The `latest` tag will be applied to the image reference when it's missing:

https://github.com/dominodatalab/distributed-compute-operator/blob/1b2286d9f24e0c715b8941b04b4d21c954dba9e6/pkg/util/util.go#L44

I can remove this default an allow an empty string to propagate but it's worth noting this will create an inconsistency between the image reference and what's in the spec, right?

```yaml
apiVersion: distributed-compute.dominodatalab.com/v1alpha1
kind: RayCluster
metadata:
  name: steve
spec:
  image:
    repository: steve-o
    # no indication of "latest"

apiVersion: apps/v1
kind: Deployment
metadata:
  name: steve-o-ray
spec:
  template:
    spec:
      containers:
      - image: steve-o:latest
```

Does that implicit bother you?

api/v1alpha1/raycluster_webhook.go
sonnysideup: I'm not sure I understand the first question. The predicate preceding this field error allows values `>= 0` so that users can scale the deployment to zero. If the HPA is present, it sees this as an [implicit maintenance-mode](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation). Am I missing something here?

Re: scale-down support, I'm going to assume you mean "HPA scale down to zero". According to the [docs](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates), this is only possible when the alpha feature gate `HPAScaleToZero` is enabled.

pkg/crd/crd.go
sonnysideup: I totally agree. I'll converge on a single pattern.

pkg/manager/manager.go
sonnysideup: Great question. I don't want users to EVER disable webhooks but I wanted to provide developers a way to disable the webhooks so they could execute the binary directly on their workstation with a debugger if necessary. I think there's probably something more clever that I can do so devs can run the entire project from their workstation but I backlogged that and added this a quick fix.

pkg/resources/ray/deployment.go
sonnysideup: You're correct. Technically, adding type metadata is only REQUIRED when using the dynamic client to create objects from unstructured data. I added this to enrich my logs during `create` operations; `update` and `delete` already have the type metadata added by the API server.

```
# with type metadata
distributed-compute-operator 2021-03-09T19:49:59.965Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"apiVersion": "v1", "kind": "ServiceAccount", "namespace": "default", "name": "example-ray"}}
distributed-compute-operator 2021-03-09T19:49:59.969Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"apiVersion": "v1", "kind": "Service", "namespace": "default", "name": "example-ray-head"}}
distributed-compute-operator 2021-03-09T19:49:59.993Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"apiVersion": "networking.k8s.io/v1", "kind": "NetworkPolicy", "namespace": "default", "name": "example-ray-cluster"}}
distributed-compute-operator 2021-03-09T19:49:59.998Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"apiVersion": "networking.k8s.io/v1", "kind": "NetworkPolicy", "namespace": "default", "name": "example-ray-client"}}
distributed-compute-operator 2021-03-09T19:50:00.003Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"apiVersion": "autoscaling/v2beta2", "kind": "HorizontalPodAutoscaler", "namespace": "default", "name": "example-ray"}}

# without type metadata
distributed-compute-operator 2021-03-09T19:49:59.965Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"namespace": "default", "name": "example-ray"}}
distributed-compute-operator 2021-03-09T19:49:59.969Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"namespace": "default", "name": "example-ray-head"}}
distributed-compute-operator 2021-03-09T19:49:59.993Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"namespace": "default", "name": "example-ray-cluster"}}
distributed-compute-operator 2021-03-09T19:49:59.998Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"namespace": "default", "name": "example-ray-client"}}
distributed-compute-operator 2021-03-09T19:50:00.003Z     INFO    controllers.RayCluster creating controlled object      {"raycluster": "default/example", "object": {"namespace": "default", "name": "example-ray"}}
```

I could remove this from the object generators and add it inside `RayClusterReconciler.createOrUpdateOwnedResource()` instead, but that's just another form of 💩. The current implementation is gross but I cannot think of any other way to accomplish this right now. You?

api/v1alpha1/raycluster_webhook.go
steved: uhyup, I misread this entirely.

api/v1alpha1/raycluster_webhook.go
steved: Yep, I was thinking of `registry` for the first one.

> I can remove this default an allow an empty string to propagate but it's worth noting this will create an inconsistency between the image reference and what's in the spec, right?

Yes, though if it implicitly changes it to `v1.2.0-cpu` that would strike me a bit odd as well. Maybe we just hard-reject empty tags given it's a pretty bad idea overall.

controllers/raycluster_controller.go
sonnysideup: So....`ray.HorizontalPodAutoscalerObjectMeta` will return an error when `rc.Spec.Autoscaling` is `nil`. If I want to follow the same pattern as the other reconcile methods, then I need to change the first func so that it will add some arbitrary, default values for `MinReplicas`, `MaxReplicas` and `AverageCPUUtilization` when the autoscaling field is not present.

I'm on the fence either way. You?

pkg/resources/ray/ray.go
sonnysideup: All object names are limited to 253 characters, include CRs. If a name is too long, the K8s API will throw an error.

e.g. using service accounts

```console
$ kubectl create sa $(ruby -e "puts 'a' * 253")
serviceaccount/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa created
$ kubectl create sa $(ruby -e "puts 'a' * 254")
The ServiceAccount "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" is invalid: metadata.name: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": must be no more than 253 characters
```

pkg/resources/ray/deployment.go
steved: if it's literally just for logging, does something like this work?
 ```
 		kinds, _, err := r.Scheme.ObjectKinds(controlled)
		if err != nil {
			return err
		}
		
		log.Info("creating controlled object", "kind", kinds[0], "object", controlled)
```

controllers/raycluster_controller.go
steved: Oh, I see, right this is the only really configurable method. I think it's fine as-is (maybe a comment).

pkg/resources/ray/deployment.go
sonnysideup: It does work, awesome find. I'm still learning how to use `runtime.Scheme`. 🙃

sonnysideup: done.

pkg/crd/crd.go
sonnysideup: done

controllers/raycluster_controller.go
sonnysideup: Ha! I was working on adding that this morning. Check out https://github.com/dominodatalab/distributed-compute-operator/pull/2/commits/1efdd6b56eb95f35355d1c7bf7eaf205ca6a4825.

deploy/helm/distributed-compute-operator/templates/hooks.yaml
sonnysideup: ack. I think what I want to do after I merge this PR is create another strictly for all of the Helm chart changes; this is more of a _template_ for the final version. I can explore using Bitnami's common chart library and address all of your other chart concerns at that time.

api/v1alpha1/raycluster_webhook.go
sonnysideup: modified accordingly

PR #16, opened 2021-03-11
Optimize helm chart

sonnysideup:
This change adds and leverages Bitnami's [common](https://github.com/bitnami/charts/tree/master/bitnami/common) chart, refactors the `image` field and reusable macros, and names global resources in a such manner that they will be tolerant of multiple releases.

Additionally,  this PR adds Helm chart publishing to CI. Here's how it works:

- The helm chart is published as part of the release process (i.e. someone pushes a semver tag)
- Helm chart `appVersion` (default image tag) and `version` (actual chart version) are set to the tag value
- Container artifacts are pushed to `ghcr.io/dominodatalab/distributed-compute-operator`
- Helm artifacts are pushed to `ghcr.io/dominodatalab/helm/distributed-compute-operator`

/cc @steved


deploy/helm/distributed-compute-operator/templates/_helpers.tpl
```diff
@@ -1,90 +1,68 @@
...
+*/}}
+{{- define "dco.webhook.annotation" -}}
+cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "dco.webhook.certificate" . }}
+{{- end -}}
+
+{{/*
+Returns a name suitable for all manager RBAC objects
+*/}}
+{{- define "dco.rbac.managerName" -}}
+dominodatalab:operator:{{ include "common.names.fullname" . }}:manager
```
steved, line 60: This might need to be a bit more "global" if we support multiple installs (since you can have the same release in two namespaces)

sonnysideup: see https://github.com/dominodatalab/distributed-compute-operator/pull/22

PR #18, opened 2021-03-21
Sets registry to blank during dev deployment

sonnysideup:
We're not building an image with the "ghcr.io" prefix so we should remove this default.

/cc @ddl-audi


saahildhulla (approved):
👍

PR #21, opened 2021-03-28
Adds support for running in Istio environments

sonnysideup:
### Changes to application

- Adds `--istio-enabled` flag to command and codifies Istio proxy readiness check + quitquitquit wrapper around CRD actions

### Changes to chart

- Adds pod security policy
- Adds network policy
- Adds new values to control resource creation/behavior depending on deployment environment


sonnysideup:
@ddl-audi @steved I just need to add some tests and/or tweak this codecov config (I know, I know, coverage is 🤮).

One of the interesting things to note about this change is the way we handle PSP permissions specific to a cluster (i.e. `RayCluster`, `SparkCluster`, etc...). If you deploy the operator into an environment with the PSP admission controller enabled and Istio available, then these changes will allow the operator/crd install to work.

If you then try to create a `RayCluster`, as an example, that might fail. It all depends  the PSP/ServiceAccount/RBAC chain establish for that cluster.

1. Someone (externally) creates a PSP for their use and then provides that `podSecurityPolicy` name in the CR. Assuming the PSP is adequate (covers crazy Istio elevated permissions shiz), then this works.
2. Someone sets up the full PSP/ServiceAccount/RBAC chain externally and provides a `serviceAccountName` in the CR. Again, if it's correctly set up, then this works.
3. Someone tries to create a CR without any of the above 2 measures in the environment I described, then 🔥🔥🔥🔥.

I think we have covered internally, correct?

Initially, I decided NOT to create PSPs for clusters since I didn't want to manage external resources via a _finalizer_. But, since we're now using StatefulSets and clean up their PVCs on CR delete, perhaps this changes the equation? I don't think it obviates my initial concern but I'm curious to get your input.

ddl-audi (dismissed):
Looks fine to me based on what we discussed

deploy/helm/distributed-compute-operator/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,50 @@
+{{- if .Values.podSecurityPolicy.enabled }}
+{{- $elevatePermissions := and .Values.istio.enabled (not .Values.istio.cniPluginInstalled) -}}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ include "common.names.fullname" . }}
```
steved, line 6: Just so I sound like a broken record, but just keep in mind this is one of those resources that should include the Helm install's namespace in the name so that multiple installs don't conflict (if it's even possible to install multiple DCOs in multiple namespaces).

(clusterrole too https://github.com/dominodatalab/distributed-compute-operator/pull/16#discussion_r594808462)

steved (commented):
> Initially, I decided NOT to create PSPs for clusters since I didn't want to manage external resources via a finalizer. But, since we're now using StatefulSets and clean up their PVCs on CR delete, perhaps this changes the equation? I don't think it obviates my initial concern but I'm curious to get your input.

I think instinctively it feels like we're fighting K8s and I'm not sure I'm a huge fan of controlling these global resources in the controller since it runs into ownership issues and such.

deploy/helm/distributed-compute-operator/templates/podsecuritypolicy.yaml
sonnysideup: I must be living in a _Helm v2_ world where namespaced releases didn't exist. I just didn't click instinctively.

If someone creates a release named `steve-o` in two separate namespaces, this and the clusterrole/binding will expose. Thanks for the catch.

PR #22, opened 2021-03-31
Jazz 🙌 support for Istio

sonnysideup:
- make webhook port configurable
- adds missing Istio resource required to make webhooks work in an environment using "STRICT mTLS"
- makes global resource names unique (👀 @steved )


ddl-audi (approved):

sonnysideup:
@steved I'm using unique names for now until I have time to revisit "global" workloads for all the Dominos.

PR #23, opened 2021-04-02
DCO Spark

ddl-audi:
Add creation of configmap representing a spark-default.conf if requested. 

- separate CM entries for master and worker pods
- generates a spark specific file and mounts in to requested location


pkg/resources/spark/service.go
```diff
@@ -59,6 +59,16 @@ func NewHeadlessService(sc *dcv1alpha1.SparkCluster) *corev1.Service {
 		Spec: corev1.ServiceSpec{
 			ClusterIP: corev1.ClusterIPNone,
 			Selector:  SelectorLabels(sc),
+			// Ports: []corev1.ServicePort{
```
ddl-audi, line 62: i need to back this out

pkg/resources/spark/statefulset.go
```diff
@@ -213,6 +244,10 @@ func componentEnvVars(sc *dcv1alpha1.SparkCluster, comp Component) []corev1.EnvV
 				Name:  "SPARK_WORKER_WEBUI_PORT",
 				Value: strconv.Itoa(int(sc.Spec.DashboardPort)),
 			},
+			{
```
ddl-audi, line 247: i wont back this out, even though its a bit unrelated. its harmless

PR #24, opened 2021-04-09
Compat changes

sonnysideup:
### Helm Chart

- relaxes `kubeVersion` constraint to allow deployment to v1.15 clusters
- supports `admissionregistration.k8s.io/v1beta1` API version so that mutating and validating webhook configurations can be used in v1.15 clusters

### Code Changes

- Generates CRDs in both `v1` and `v1beta1` formats
- CRD apply/delete logic branches based on retrieved K8s version (i.e. if version is >= 1.1.6, use V1 API, else use V1Beta1)

### DIFF
--- a/pkg/crd/v1beta1.go
+++ b/pkg/crd/v1beta1.go
@@ -1,0 +1,128 @@
+package crd
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+
+	"github.com/blang/semver"
+	apixv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apixv1beta1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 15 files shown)


sonnysideup:
@steved @ddl-audi 

I tested this against two local clusters, v1.15.12 and v 1.16.15, and the chart/code switching works as expected. My expectation is that we'll remove support for v1.15 in the near future and, of course, I'm happy to help push this forward in any way possible.

Re: code changes, the `k8s.io/apiextensions-apiserver` pkg defines separate interfaces/clients/types for the different versions. This makes it difficult to create reusable functions that leverage a single interface. 😞 Instead, I create a separate set of duplicate, "v1beta1" functions and tried to limit my modifications to the existing code paths (because of my formerly-stated expectation). It's not DRY, I can't look at myself in the mirror right now, but this works.

pkg/crd/v1beta1.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return
+	}
+
+	version, err := semver.Parse(strings.TrimPrefix(info.GitVersion, "v"))
+	if err != nil {
+		return
+	}
+
+	return version.GTE(v1VersionConstraint), nil
```
steved, line 77: Doesn't `version.Info` have all this split out info?

steved (approved):

PR #29, opened 2021-05-18
helm chart: support priority class name setting

steved:
Allow cluster-wide pod priority & preemption settings (though it likely won't impact DCO much).


sonnysideup (approved):
@steved, are we supporting priority classes now? I guess the answer should be obvious.

PR #38, opened 2021-07-01
Adding dominodatalab.com labels to Service, NetworkPolicy to Spark DCO

saahildhulla:
Ensure the correct `dominodatalab.com` labels exist on the correct K8s resources for DCO Spark

### DIFF
--- a/pkg/resources/spark/networkpolicy.go
+++ b/pkg/resources/spark/networkpolicy.go
@@ -27,7 +27,7 @@ func NewClusterWorkerNetworkPolicy(sc *dcv1alpha1.SparkCluster) *networkingv1.Ne
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      InstanceObjectName(sc.Name, ComponentWorker),
 			Namespace: sc.Namespace,
-			Labels:    MetadataLabels(sc),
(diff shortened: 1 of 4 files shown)


pkg/resources/spark/networkpolicy.go
```diff
@@ -73,7 +73,7 @@ func NewClusterDriverNetworkPolicy(sc *dcv1alpha1.SparkCluster) *networkingv1.Ne
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      InstanceObjectName(sc.Name, "driver"),
 			Namespace: sc.Namespace,
-			Labels:    MetadataLabels(sc),
+			Labels:    MetadataAndAdditionalLabels(MetadataLabels(sc), sc.Spec.Master.Labels),
```
saahildhulla, line 76: in theory it doesn't matter which set of extra labels (worker/master) are added onto the `MetadataLabels`, they both contain the same `dominodatalab.com` labels

pocheung1:
I think we should use the same approach as Dask to add global labels but I will defer to @sonnysideup on this.
https://github.com/dominodatalab/distributed-compute-operator/blob/2e68bab1794d1075f34e216aeeaf0082719de5f4/pkg/cluster/dask/metadata.go#L19

PR #39, opened 2021-07-01
Adding global labels to DCO Spark

saahildhulla:
Spark services and network policies are missing `dominodatalab.com` labels.


sonnysideup (commented):
@saahildhulla what about all the other resources, like network policy, rbac, hpas, etc...?

saahildhulla:
@sonnysideup good call, the bugfix ticket only mentioned services and netpols, but it can't hurt having all affiliated K8s resources to have those labels. Added as a part of the latest commit.

sonnysideup (approved):
🚢 LGTM

PR #43, opened 2021-07-02
Fix Autoscaling AverageCPUUtilization

saahildhulla:
Autoscaling `AverageCPUUtilization` was previously an optional value, and had been made mandatory through recent changes. This fix converts it back into an optional value.

Tested against my local.


pocheung1 (approved):
LGTM

PR #46, opened 2021-07-29
Move KeyTab config to Spec Level

saahildhulla:
Need to move KeyTab config into the Spec level instead of setting it on master/workers directly.


config/samples/distributed-compute_v1alpha1_sparkcluster.yaml
```diff
@@ -59,7 +59,7 @@ spec:
 
   # KeyTab Config for Kerberos
   #  kerberosKeytab:
-  #    path: /tmp/keytab.conf
+  #    mountPath: /tmp/keytab.conf
   #    configs: dGVzdHZhbHVlCg==
```
pocheung1, line 63: `configs` should be renamed to `contents`?

saahildhulla: yep, my mistake!

api/v1alpha1/sparkcluster_types.go
```diff
@@ -137,6 +127,9 @@ type SparkClusterSpec struct {
 
 	// Autoscaling parameters used to scale up/down spark worker nodes.
 	Autoscaling *Autoscaling `json:"autoscaling,omitempty"`
+
+	// KerberosKeytab configures the Kerberos Keytab for Spark
+	KerberosKeytab *KerberosKeytabConfig `json:"kerberosKeytab,omitempty"`
```
sonnysideup, line 132: 👍

api/v1alpha1/sparkcluster_webhook.go
```diff
@@ -252,11 +252,7 @@ func (sc *SparkCluster) validateFrameworkConfigs() field.ErrorList {
 func (sc *SparkCluster) validateKeyTabConfigs() field.ErrorList {
 	var errs field.ErrorList
 
-	if err := sc.validateKeyTabConfig(sc.Spec.Master.KeyTabConfig, "master"); err != nil {
-		errs = append(errs, err...)
-	}
-
-	if err := sc.validateKeyTabConfig(sc.Spec.Worker.KeyTabConfig, "worker"); err != nil {
+	if err := sc.validateKeyTabConfig(sc.Spec.KerberosKeytab); err != nil {
```
sonnysideup, line 255: Is it possible to use my _shared validation func_ here?

https://github.com/dominodatalab/distributed-compute-operator/blob/3aeef3d7a76c3b4f3e21cd36b7ca38ad62c1dc86/api/v1alpha1/validations.go#L153

pkg/resources/spark/configmap.go
```diff
@@ -36,15 +36,16 @@ func NewFrameworkConfigMap(sc *dcv1alpha1.SparkCluster) *corev1.ConfigMap {
...
 	binaryData := map[string][]byte{}
-	if sc.Spec.Master.KeyTabConfig != nil {
-		binaryData[string(ComponentMaster)] = sc.Spec.Master.KeyTabConfig.KeyTab
-	}
-	if sc.Spec.Worker.KeyTabConfig != nil {
-		binaryData[string(ComponentWorker)] = sc.Spec.Worker.KeyTabConfig.KeyTab
+
+	if sc.Spec.KerberosKeytab != nil {
+		binaryData[string(ComponentMaster)] = sc.Spec.KerberosKeytab.Contents
+		binaryData[string(ComponentWorker)] = sc.Spec.KerberosKeytab.Contents
```
sonnysideup, line 42: I just wonder if we have to maintain two separate keys for master/worker with the same keytab infos? Should we condense this into a single field (e.g. `"keytab": []byte("dem contents")`  and remove these extra bits?

See https://github.com/dominodatalab/distributed-compute-operator/blob/3aeef3d7a76c3b4f3e21cd36b7ca38ad62c1dc86/pkg/cluster/dask/configmap.go#L36

sonnysideup (approved):
LGTM. No major but check out my comments regarding a couple of refactoring questions/suggestions.

PR #49, opened 2021-08-10
Disable TCP Idle Timeout for Istio

saahildhulla:
Istio kills idle TCP connections after 1 hour. We need to disable this time for long-running Spark jobs.

Adding an`EnvoyFilter` resource to disable the timeout for INBOUND and OUTBOUND traffic. After much research, this seemed like the most straightforward way to accomplish what we want. The `EnvoyFilter` is deployed with the DCO, and deployed on a per-SparkCluster basis. Seeing as the folks using this would be using DCO/Spark with Istio v1.10 on a Domino 4.6.0 or later deployment, this seemed like a safe assumption/decision to make.

Tested locally in an Istio-enabled Minikube environment and against a deploy. 

Confirmed that the timeout was propagated through to the `istio-proxy` by running `istioctl proxy-config listeners [example-spark-master-0/example-spark-worker-X] -o json`

Also lumping in adding global labels to RayCluster PVC's, when they are created.


pkg/manager/manager.go
```diff
@@ -81,9 +81,10 @@ func Start(cfg *Config) error {
 	}
 
 	if err = (&controllers.SparkClusterReconciler{
-		Client: mgr.GetClient(),
-		Log:    logging.New(ctrl.Log.WithName("controllers").WithName("SparkCluster")),
-		Scheme: mgr.GetScheme(),
+		Client:       mgr.GetClient(),
+		Log:          logging.New(ctrl.Log.WithName("controllers").WithName("SparkCluster")),
+		Scheme:       mgr.GetScheme(),
+		IstioEnabled: cfg.IstioEnabled,
```
saahildhulla, line 87: this field was previously unspecified (I found a bug :) )

pkg/resources/spark/statefulset.go
```diff
@@ -14,11 +14,14 @@ import (
 	"github.com/dominodatalab/distributed-compute-operator/pkg/util"
 )
 
-const frameworkConfigMountPath = "/opt/bitnami/spark/conf/spark-defaults.conf"
+const (
+	frameworkConfigMountPath = "/opt/bitnami/spark/conf/spark-defaults.conf"
+	istioProxyMetadata       = "    proxyMetadata:\n      ISTIO_META_IDLE_TIMEOUT: \"0s\""
```
sonnysideup, line 19: I think I'd prefer us to use a YAML lib to apply YAML formatting.

```go
import "sigs.k8s.io/yaml"

var (
	istioProxyConfig = map[string]map[string]string{
		"proxyMetadata": {
			"ISTIO_META_IDLE_TIMEOUT": "0s",
		},
	}
	istioProxyConfigAnnotationKey   = "proxy.istio.io/config"
	istioProxyConfigAnnotationValue string
)

// put at the end of the file
func init() {
	bs, err := yaml.Marshal(istioProxyConfig)
	if err != nil {
		panic(fmt.Errorf("cannot marshal istio proxy config: %w", err))
	}

	istioProxyConfigAnnotationValue = string(bs)
}
```

deploy/helm/distributed-compute-operator/templates/istio.yaml
```diff
@@ -14,4 +14,27 @@ spec:
   portLevelMtls:
     {{ .Values.config.webhookPort }}:
       mode: PERMISSIVE
+---
+## Disables the Istio TCP idle timeout -- required for SparkClusters
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
```
sonnysideup, line 20: Doesn't this disable ALL sidecar timeouts within the namespace where it's deployed? It should ONLY target Spark clusters and nothing else.

I'm not sure this is the right place to create this resource since it doesn't provide you with programmatic access, right?

sonnysideup (commented):
If the annotation does not handle INBOUND traffic, can the `EnvoyFilter` handle both so that we don't have to _spread_ idle timeout configuration across two locations?

config/samples/distributed-compute_v1alpha1_sparkcluster.yaml
```diff
@@ -104,7 +104,7 @@ spec:
   #    scaleDownStabilizationWindowSeconds: 10
 
   # override istio's peer authentication policy
-  #  istioMutualTLSMode: PERMISSIVE
+  istioMutualTLSMode: STRICT
```
sonnysideup: did you mean to change this template?

api/v1alpha1/sparkcluster_types.go
```diff
@@ -79,6 +79,8 @@ type SparkClusterSpec struct {
 
 	GlobalLabels map[string]string `json:"globalLabels,omitempty"`
 
+	EnvoyFilterLabels map[string]string `json:"envoyFilterLabels,omitempty"`
```
sonnysideup, line 82: Maybe add a comment here? this will propagate into the CRD and be availabe for trying to figure out how this field works.

It _seems_ fine to add this here vs embedding. We can group this with a spark-specific istio config in the future and embed that if we want. The front-facing API will probably stay the same.

pkg/resources/ray/statefulset.go
```diff
@@ -350,7 +352,8 @@ func processPVCTemplates(vcts []dcv1alpha1.PersistentVolumeClaimTemplate) (pvcTm
 
 		pvcTmpls = append(pvcTmpls, corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: vct.Name,
+				Name:   vct.Name,
+				Labels: rc.Spec.GlobalLabels,
```
sonnysideup, line 356: 👍

pkg/resources/spark/envoyfilter.go
```diff
@@ -0,0 +1,90 @@
...
+	"istio.io/api/networking/v1alpha3"
+	v1alpha32 "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	dcv1alpha1 "github.com/dominodatalab/distributed-compute-operator/api/v1alpha1"
+)
+
+const (
+	filterName = "envoy.filters.network.tcp_proxy"
+)
```
sonnysideup, line 16:
```suggestion
const envoyFilterName = "envoy.filters.network.tcp_proxy"
```

pkg/resources/spark/envoyfilter.go
```diff
@@ -0,0 +1,90 @@
...
+
+// NewEnvoyFilter creates a new EnvoyFilter resource to set idle_timeout for Istio-enabled deployments
+func NewEnvoyFilter(sc *dcv1alpha1.SparkCluster) (v1alpha32.EnvoyFilter, error) {
```
sonnysideup, line 19:
```suggestion
func NewEnvoyFilter(sc *dcv1alpha1.SparkCluster) *v1alpha32.EnvoyFilter {
```

You don't need to return an `error` since there's nothing in this func that can create one, right? Also, you may want to return a pointer, no?

pkg/resources/spark/envoyfilter.go
```diff
@@ -0,0 +1,90 @@
+package spark
+
+import (
+	"fmt"
+
+	protobuftypes "github.com/gogo/protobuf/types"
+	"istio.io/api/networking/v1alpha3"
+	v1alpha32 "istio.io/client-go/pkg/apis/networking/v1alpha3"
```
sonnysideup, line 8:
```suggestion
	networkingv1alpha3 "istio.io/api/networking/v1alpha3"
	apinetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
```

disambiguate?

pkg/resources/spark/envoyfilter_test.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"testing"
+
+	protobuftypes "github.com/gogo/protobuf/types"
+	v1alpha32 "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"istio.io/api/networking/v1alpha3"
```
sonnysideup, line 12: needs to be sorted

pkg/resources/spark/statefulset.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/dominodatalab/distributed-compute-operator/pkg/util"
 )
 
-const frameworkConfigMountPath = "/opt/bitnami/spark/conf/spark-defaults.conf"
+const (
+	frameworkConfigMountPath = "/opt/bitnami/spark/conf/spark-defaults.conf"
+)
```
sonnysideup, line 19:
```suggestion
const frameworkConfigMountPath = "/opt/bitnami/spark/conf/spark-defaults.conf"
```

scripts/development.sh
```diff
@@ -122,7 +122,8 @@ function dco::helm_install() {
     --set image.registry="" \
     --set image.repository="$IMAGE_NAME" \
     --set image.tag="$latest_tag" \
-    --set config.logDevelopmentMode=true
+    --set config.logDevelopmentMode=true \
+    --set istio.enabled=true
```
sonnysideup: I guess we want to "default" to using Istio? This doesn't seem like a crazy default, but I want to make sure this was your intention.

pkg/resources/spark/statefulset.go
```diff
@@ -123,7 +128,7 @@ func NewStatefulSet(sc *dcv1alpha1.SparkCluster, comp Component) (*appsv1.Statef
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Labels:      labels,
+					Labels:      util.MergeStringMaps(labels, sc.Spec.EnvoyFilterLabels),
```
sonnysideup:
```suggestion
					Labels:      util.MergeStringMaps(sc.Spec.EnvoyFilterLabels, labels),
```

pkg/util/util_test.go
```diff
@@ -54,6 +54,23 @@ func TestMergeStringMaps(t *testing.T) {
 	assert.Equal(t, expected, actual)
 }
 
+func TestMergeStringMapsWithEmpty(t *testing.T) {
```
sonnysideup, line 57: subtest in `TestMergeStringMaps` (i.e. `t.Run("empty_src", ...`)?

api/v1alpha1/sparkcluster_types.go
saahildhulla: I agree.

sonnysideup (approved):

PR #52, opened 2021-08-12
Using apiextensions v1 API exclusively

sonnysideup:
v1beta1 support will be removed from k8s v1.22 and v1 has been supported
since k8s v1.16. this removes support for v1beta1 CRDs and, by
extension, means we can start leveraging "version conversion". this will
allow us to modify our CRD APIs and maintain backwards compatibility
with earlier versions.


saahildhulla (approved):
lgtm!

PR #53, opened 2021-08-16
Allows dco to process CRs in other namespaces

sonnysideup:
This does not require the application to be re-deployed but does allow us the ability the move the DCO into a different namespace. I think that operationally this design makes more sense since DCO is more of a cluster component.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -10,7 +10,6 @@ import (
 )
 
 var (
-	namespace            string
 	probeAddr            string
 	metricsAddr          string
 	webhookPort          int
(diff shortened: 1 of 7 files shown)


saahildhulla (approved):
This makes sense, but I have one question: if/when DCO moves to a separate namespace, will there be some namespace specification that needs to be provided on compute clusters when they are launched? Ex. DCO lives in `namespace-a` but launch/monitor/reconcile SparkCluster in `namespace-z`?

sonnysideup:
https://github.com/dominodatalab/distributed-compute-operator/pull/53#pullrequestreview-731054682

@saahildhulla DCO will now watch ALL namespaces without any extra specifications. As an example, if DCO was running in and created a spark cluster in `ns-a`, after moving it to `ns-b`, then it will still monitor/reconcile those clusters in `ns-a`.

I decided to add https://github.com/dominodatalab/distributed-compute-operator/pull/53/commits/77b53fdbc5da6cf153a1eba47187a3717f5ea1c0 so that one _could limit the watched namespaces_.

PR #55, opened 2021-08-29
Ray Refactor - Part 1 (API config migration)

sonnysideup:
- Leverages common configuration types created for Dask API (these are common to all cluster types)
- Makes changes to sample config
- Changes controller & resource generators to accommodate type changes
- Updates to go 1.17


sonnysideup:
@saahildhulla These changes ONLY modify the API types used to create Ray clusters but leave the old `pkg/resources/ray` bits in place for now. After this is merged, I can build reusable components for each cluster implementation by using the same _common types_ defined earlier for Dask.

@pocheung1 There are some minor changes to the API but they match the Dask API. Namely,

```yaml
spec:
  # control the creation of network policies
  # networkPolicy:
  #   enabled: true
  #   clientServerLabels:
  #     ray-client: "true"
  #   dashboardLabels:
  #     ray-dashboard-client: "true"

  # specify the service account used to run ray pods
  # serviceAccountName: my-service-account
```

becomes

```yaml
spec:
  # networkPolicy:
  #   enabled: true
  #   clientLabels: {}
  #   dashboardLabels: {}

  # serviceAccount:
  #   name: ""
  #   automountServiceAccountToken: false
```

This _mirrors_ the 95% of Dask API, excluding Ray-specifc ports and memory settings. See https://github.com/dominodatalab/distributed-compute-operator/blob/ray-refactor/config/samples/distributed-compute_v1alpha1_raycluster.yaml for the new, full template.

pocheung1:
@sonnysideup thanks for the heads up.  Created https://dominodatalab.atlassian.net/browse/PLAT-2831 to update cluster configuration in Scala.

saahildhulla (approved):
LGTM

pocheung1:
@sonnysideup Do you plan to update Spark to use the new ServiceAccount spec?  Would it make sense to bundle the changes together?

sonnysideup:
> @sonnysideup Do you plan to update Spark to use the new ServiceAccount spec? Would it make sense to bundle the changes together?

@pocheung1 I can add that update as well. That would converge all 3 cluster types (Dask, Ray, and Spark) so that their APIs are BnL the same. I'd probably create a separate PR for Spark but I can add those changes to this one if it makes it easier for you to test. What are your thoughts?

pocheung1:
That would be great, thanks @sonnysideup.   Either way is fine.  It's up to you.

pocheung1 (approved):

PR #56, opened 2021-09-01
Spark Refactor - Part 1 (API config migration) 

sonnysideup:
- Leverages common configuration types created for Dask API (these are common to all cluster types)
- Makes changes to sample config
- Changes controller & resource generators to accommodate type changes

@pocheung1 I trimmed the API fields a considerable amount. The same changes that were made to Ray have been made here, but there are some additional changes worth highlighting:

1. I removed the TCP prefix from `TCPMasterWebPort` and `TCPWorkerWebPort` because it's not relevant (https://github.com/dominodatalab/distributed-compute-operator/pull/56/files#diff-eacc13f76cb9b7221a6f1b2e0c25046da3322b0696a679ea9c0d3f3575de6eddR60-R62)
2. I removed `DashboardPort` and `EnableDashboard` because (a) there's no way to disable the dashboard and (b) the dashboard is accessible via the `MasterWebPort`.
3. The driver labels should be added to `spec.networkPolicy.clientLabels`
4. `spark-defaults.conf` can be configured via `spec.(master|worker).defaultConfiguration` fields
5. Driver settings are now located at `spec.driver` and only require you to specify ports and a pod selector (i.e. you do not need to name the ports, that's our job since we manage the service).

I think that's it but you can always check [distributed-compute_v1alpha1_sparkcluster.yaml ](https://github.com/dominodatalab/distributed-compute-operator/blob/99115fc19569955b2cc72802da843d3dce8d6937/config/samples/distributed-compute_v1alpha1_sparkcluster.yaml) for an example of the new spec. A lot has changed so feel free to reach out if you want to review these changes a little deeper.


pocheung1:
@sonnysideup I updated the DCO image to `pr-56` on my deploy, applied [distributed-compute_v1alpha1_sparkcluster.yaml](https://raw.githubusercontent.com/dominodatalab/distributed-compute-operator/99115fc19569955b2cc72802da843d3dce8d6937/config/samples/distributed-compute_v1alpha1_sparkcluster.yaml) and got:

```
$ k apply -f distributed-compute_v1alpha1_sparkcluster.yaml
error: error validating "distributed-compute_v1alpha1_sparkcluster.yaml": error validating data: [ValidationError(SparkCluster.spec): unknown field "driver" in com.dominodatalab.distributed-compute.v1alpha1.SparkCluster.spec, ValidationError(SparkCluster.spec): unknown field "workerMemoryLimit" in com.dominodatalab.distributed-compute.v1alpha1.SparkCluster.spec]; if you choose to ignore these errors, turn validation off with --validate=false
```

sonnysideup:
https://github.com/dominodatalab/distributed-compute-operator/pull/56#issuecomment-911203067

@pocheung1 I _think_ this issue might be that you updated the image but didn't upgrade the chart. Is that right? Without an upgrade to the chart, then the CRDs (which define the new fields) are not applied to the cluster. I would expect this kind of error in that case.

Let's chat sometime today.

pocheung1:
Thanks @sonnysideup.  That was my bad.  Sorry for the false alarm.

saahildhulla (approved):

saahildhulla:
I didn't have much of a hand in this PR set on either side, so as long as y'all both @sonnysideup @pocheung1 are in sync then I have no problems! 👍

pocheung1 (approved):

PR #57, opened 2021-09-16
Ray+PyTorch+Istio Support

sonnysideup:
### Changes

Whenever Istio is enabled, Ray Statefulsets will use specific annotations to explicitly configure envoy so that only well-defined ports are redirected through the sidecar proxy. this is necessary because certain ML frameworks, like PyTorch, use
random ports and odd protocols to distribute work. Inter-cluster networking fails without this type of configuration for these cases.

The specific ports added to the annotation are the same that have been added to the `Service` with protocol prefixes so STRICT mTLS _would work_ even without this configuration.


saahildhulla (approved):
looks good to me 👍

PR #58, opened 2021-09-29
allow retries of CRD hooks during helm install

steved:
There is the chance of a race condition with Istio/CNI and pods on a new
node that, for a Job with a backoffLimit=0, could cause a failure when a
a simple retry would work.

### DIFF
--- a/deploy/helm/distributed-compute-operator/templates/hooks.yaml
+++ b/deploy/helm/distributed-compute-operator/templates/hooks.yaml
@@ -69,7 +69,7 @@ metadata:
     "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
-  backoffLimit: 0
+  backoffLimit: 3
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


deploy/helm/distributed-compute-operator/templates/hooks.yaml
```diff
@@ -69,7 +69,7 @@ metadata:
     "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
-  backoffLimit: 0
+  backoffLimit: 3
```
sonnysideup, line 72: Any reason to set this to 3 instead of removing the limitation and using K8s' default value? I think it's 6.

sonnysideup (approved):

deploy/helm/distributed-compute-operator/templates/hooks.yaml
steved: Guess not, nope. Updated.

sonnysideup (approved):

PR #59, opened 2021-10-11
MPI job support

sonnysideup:
## Included

- New `MPIJob` API w/ defaulting/validating
- Controller that will launch training job via `mpirun`

## Missing

- Support for non-root execution
- Unit tests
- Disallow certain, if not all, updates to `spec` (Job execution is weird enough that we want to restrict what can be modified after creation or, alternatively, recreate the Job?)
- Istio support


adp312:
The MPI work continues based on renewed requirements. This PR is closed and archived.

PR #64, opened 2022-01-05
DOM-35162 - Remove launcher pod, rename mpijob to mpicluster

adp312:
I see now that the rename has messed everything up, as it affected many files. I should have done it in a separate commit... sorry...

The essential changes are in `pkg/cluster/mpi` directory. Specifically, `statusupdate.go` contains a new logic of tracking status of the cluster.


pkg/cluster/mpi/statusupdate.go
```diff
@@ -105,3 +98,20 @@ func (c statusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
+	objKey := client.ObjectKey{
+		Name:      serviceName(cr),
+		Namespace: cr.Namespace,
+	}
+
+	if err := ctx.Client.Get(ctx, objKey, &ep); err != nil && !apierrors.IsNotFound(err) {
+		return 0, fmt.Errorf("cannot fetch endpoints: %w", err)
+	}
+
+	if len(ep.Subsets) == 1 {
```
ddl-kevin, line 113: This seems a little brittle, I'm not sure how this works.

pkg/cluster/mpi/statusupdate.go
```diff
@@ -105,3 +98,20 @@ func (c statusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
 }
+
+func getActivePodCnt(ctx *core.Context, cr *dcv1alpha1.MPICluster) (int, error) {
+	var ep corev1.Endpoints
+	objKey := client.ObjectKey{
+		Name:      serviceName(cr),
+		Namespace: cr.Namespace,
+	}
+
+	if err := ctx.Client.Get(ctx, objKey, &ep); err != nil && !apierrors.IsNotFound(err) {
```
ddl-kevin, line 109: Do you know if this makes an API call? Ideally this is only querying a local cache.

ddl-kevin (approved):
LGTM, I have some questions on how we determine activePodCnt.

pkg/cluster/mpi/statusupdate.go
adp312: This is really a copy-paste of Sonny's code from the portion that I have removed. Agree that this logic seems somewhat fragile. I will look at what's going on behind the scene here.

PR #65, opened 2022-01-06
DOM-35203 - Remove secret generation

adp312:


api/v1alpha1/validations.go
```diff
@@ -176,6 +176,16 @@ func validateKerberosKeytab(keytab *KerberosKeytabConfig) field.ErrorList {
 	return errs
 }
 
+func validateSharedSSHSecret(secret string) field.ErrorList {
+	var errs field.ErrorList
+	fp := field.NewPath("spec", "workers", "sharedSSHSecret")
+	if secret == "" {
```
ddl-kevin, line 182: should we try to validate that the secret exists (using K8s API?), or maybe its better for the cluster to be "valid" but get `Starting` dues to Pods not being able to mount the secret. This is how Pods behave.

ddl-kevin: It could be worth validating the "type" of secret, to ensure the key we reference in the volumeMount is present.

ddl-kevin (approved):
LGTM

api/v1alpha1/validations.go
adp312: Yes, I agree that it's worth validating that the specifying secret exists. Will do.
I am not sure about your second point... Do you mean testing that the secret's type is `kubernetes.io/ssh-auth`? Interestingly, in my tests I was able to successfully use a secret of the `Opaque` type.

ddl-kevin: Did the `Opaque` secret have the correct key (ssh-publickey)? If so, thats not surprising, but I think we should require the secret to be `kubernetes.io/ssh-auth` so that we can be sure that key exists (maybe validate that the key is present too).

adp312: Yes, this was a secret of type `Opaque` with a two `ssh-*` fields/keys.

adp312: I doesn't seem like I can easily check presence of a secret + key in that validator function because it lives in a webhook that doesn't have access to the k8s context (or, I am sure it's possible to do, but with some extra hassle and haven't been done for anything else, as far as I can tell). What I will do is an additional checking inside the CR reconciler, so that it returns a predictable error with a reasonable message. This is not ideal, but Sonny is already doing exactly that for things like image not being parseable.

ddl-kevin: Oh, I see. Interesting. I think we should just not check and let the pods enter a crashloop backoff and reflect that in the cluster status as we would for any other reason.

adp312: Yeah... Let me see what we'll get in the logs in that scenario.

adp312: It works terribly. On an invalid secret, the reconciler reports: `"cannot reconcile service: Service \"example-mpi-worker\" is invalid: metadata.resourceVersion: Invalid value: \"\": must be specified for an update"` and stops. To make things worse, the error is not reflected in the cluster status (shows "Pending"). 

I will implements an explicit secret check here, and address the proper error status reporting in DOM-35253.

ddl-kevin: Hmmm, that doesn't sound right. That sounds like the DCO is trying to update the Service resource incorrectly. I would expect to just see Pods stuck and not registered with the Service

adp312: Pods _are_ stuck.

ddl-kevin: Ok, but that error seems to be related to updating the Service resource. I wouldn't expect that.

adp312: I see.

adp312: Addressed by my changes in `pkg/cluster/mpi/statefulset.go`

PR #69, opened 2022-01-10
DOM-35418: Bug in CR provisioning and update

adp312:


pkg/controller/actions/actions.go
```diff
@@ -31,8 +44,10 @@ func CreateOrUpdateOwnedResource(ctx *core.Context, owner metav1.Object, control
 		return nil
 	}
 
-	ctx.Log.V(1).Info("Applying patch to object", "gvk", gvk, "object", controlled, "patch", string(patchResult.Patch))
-	if err = ctx.Patch.Annotator.SetLastAppliedAnnotation(controlled); err != nil {
+	ctx.Log.V(1).Info("Applying patch to object", "gvk", gvk, "object", controlled, "patch",
+		string(patchResult.Patch))
+	err = ctx.Patch.Annotator.SetLastAppliedAnnotation(controlled)
```
adp312, line 49: Not sure

pkg/controller/actions/actions.go
```diff
@@ -95,34 +111,18 @@ func DeleteStorage(ctx *core.Context, opts []client.ListOption) error {
...
-		if err = ctx.Patch.Annotator.SetLastAppliedAnnotation(controlled); err != nil {
-			return
-		}
+	return gvks[0], nil
+}
 
-		ctx.Log.V(1).Info("Creating controlled object", "gvk", gvk, "object", controlled)
-		err = ctx.Client.Create(ctx, controlled)
+func createOwnedResource(ctx *core.Context, controlled client.Object) error {
+	err := ctx.Patch.Annotator.SetLastAppliedAnnotation(controlled)
```
adp312, line 123: Not sure

pkg/controller/actions/actions.go
```diff
@@ -1,25 +1,38 @@
...
+	if err != nil {
+		return err
+	}
+
+	gvk, err := getObjectKind(ctx, controlled)
+	if err != nil {
+		return err
+	}
+
+	found := controlled.DeepCopyObject().(client.Object)
```
ddl-kevin, line 28: What does this syntax do? `DeepCopyObject().(client.Object)`

Is `found` the right name for this variable? It looks like we perform the actual check on the next line.

adp312:
The lint warning is already fixed in a separate commit. I will rebase accordingly when the time comes.

pkg/controller/actions/actions.go
ddl-kevin: This isnt clear to me either, shouldnt kube-apiserver populate this field. TBH I dont understand the new apply mechanism that this field relates to :(. 

@sonnysideup ?

ddl-kevin (approved):
This looks reasonable. Was there a reason you went this route vs. just reverting the changes?

pkg/controller/actions/actions.go
adp312: - The deepCopy method (it's auto-generated) creates a "true" copy of a the data object including proper cloning of all nested/constituent objects. I haven't looked at the details of how this is used here specifically, but that's a normally accepted practice.
- `found` is likely not a right now. It's my copy/paste of Sonny's code (a bad excuse).

pkg/controller/actions/actions.go
adp312: Never mind, then. I will separately look at what's going on there. This entire piece of code will need some final refactoring, anyway.

adp312:
Please note that the branch name here doesn't make sense. It should be something like "petrov.DOM-35418.service-update-error"

PR #70, opened 2022-01-11
DOM-35253 Refactoring of status transitions in MPI

adp312:


pkg/cluster/mpi/statusupdate.go
```diff
@@ -44,44 +47,39 @@ func (c statusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
-	if lErr := ctx.Client.List(ctx, podList, listOpts...); lErr != nil {
-		return ctrl.Result{}, fmt.Errorf("cannot list cluster pods: %w", lErr)
-	}
-
 	var podNames []string
-	for _, pod := range podList.Items {
+	var runningPodCnt = 0
+	for _, pod := range pods {
 		podNames = append(podNames, pod.Name)
+		if pod.Status.Phase == corev1.PodRunning {
```
JoshBroomberg, line 58: In the current nucleus code, we only treat the master as Ready if phase is ready and the pod condition ready is also true (I believe a pod goes to phase ready as soon as it is scheduled but only to the pod condition when it is up and healthy) or something like that

Is the choice to use only phase and not condition intentional?

pkg/cluster/mpi/statusupdate.go
```diff
@@ -19,8 +21,10 @@ import (
 )
 
 const (
-	ReadyStatus   = "Ready"
-	PendingStatus = "Pending"
+	PendingStatus  v1.JobConditionType = "Pending"
+	StartingStatus v1.JobConditionType = "Starting"
+	RunningStatus  v1.JobConditionType = "Running"
+	StoppingStatus v1.JobConditionType = "Stopping"
```
JoshBroomberg, line 27: These are in line with my code

pkg/cluster/mpi/statusupdate.go
adp312: That's a good point. I didn't use conditions because line 58 above is already an improvement over the previous logic, and I though checking something else will be an overkill. After Pod is scheduled it goes to Pending first, and only after that to Ready. So, the current logic should be _almost_ OK. But I will add checking for conditions too. 

Q: Does Nucleus check for conditions on Pod or Container? (I believe it's possible to do either way).

pkg/cluster/mpi/statefulset.go
```diff
@@ -118,14 +118,29 @@ func (c statefulSetComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
 		client.InNamespace(cr.Namespace),
 		client.MatchingLabels(meta.MatchLabels(cr)),
 	}
-
 	err := actions.DeleteStorage(ctx, pvcListOpts)
+	if err != nil {
+		return ctrl.Result{}, false, fmt.Errorf("cannot delete storage: %w", err)
+	}
+
+	sts := &appsv1.StatefulSet{
```
ddl-kevin, line 131: I guess this was getting deleted due to the owner reference before?

pkg/cluster/mpi/statefulset.go
```diff
@@ -118,14 +118,29 @@ func (c statefulSetComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
 	cr := objToMPICluster(ctx.Object)
+
 	pvcListOpts := []client.ListOption{
 		client.InNamespace(cr.Namespace),
 		client.MatchingLabels(meta.MatchLabels(cr)),
 	}
-
 	err := actions.DeleteStorage(ctx, pvcListOpts)
+	if err != nil {
+		return ctrl.Result{}, false, fmt.Errorf("cannot delete storage: %w", err)
```
ddl-kevin, line 128: So if this fails, does the CR just stay stuck deleting? Will it retry at some point?

pkg/cluster/mpi/statusupdate.go
ddl-kevin: IMO we should determine what to check to confirm that the readiness / liveness checks are passing for the Pod. I believe that might be checking that the Pods "Ready" condition is `true`. https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

pkg/cluster/mpi/statusupdate.go
```diff
@@ -91,28 +89,43 @@ func (c statusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
+		cr.Status.ClusterStatus = StoppingStatus
+		err := ctx.Client.Status().Update(ctx, cr)
+		if err != nil {
+			return ctrl.Result{}, false, fmt.Errorf("cannot update cluster status: %w", err)
+		}
 	}
 
-	if err := ctx.Client.Get(ctx, objKey, &ep); err != nil && !apierrors.IsNotFound(err) {
-		return 0, fmt.Errorf("cannot fetch endpoints: %w", err)
+	pods, err := getPods(ctx, cr)
```
ddl-kevin, line 111: What is the point of this check? I think if the StatefulSet is gone, then the Pods can be assumed to be gone.

pkg/cluster/mpi/statefulset.go
```diff
@@ -118,14 +118,29 @@ func (c statefulSetComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
+	}
+
+	sts := &appsv1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      workerStatefulSetName(cr),
+			Namespace: cr.Namespace,
+			Labels:    meta.StandardLabelsWithComponent(cr, ComponentWorker, cr.Spec.Worker.Labels),
+		},
+	}
+	err = actions.DeleteIfExists(ctx, sts)
```
ddl-kevin, line 138: Deleting is not synchronous, (because of finalizers), do you know if that is handled here?

ddl-kevin (commented):
Looking good, just a few questions.

pkg/cluster/mpi/statefulset.go
adp312: I suppose so. It's a valid question, but it's out of scope of this specific piece of code. I mean, this function obviously expects an error to be returned if something happened. Then however calls it (Sonny's custom reconciler) is supposed to do something reasonable. I believe it does, but I have never tried this sort of things explicitly. 
I will verify what's going on.

pkg/cluster/mpi/statefulset.go
adp312: This data structure is just a "stub" used to point to a service that will be deleted.

pkg/cluster/mpi/statusupdate.go
adp312: I agree. I think you both (@JoshBroomberg, @ddl-kevin) are talking about the same thing. It's an easy change.

pkg/cluster/mpi/statusupdate.go
adp312: You mean, check on the stateful set, not the pods? Yeah, perhaps. It's probably a more correct way of getting a result, but on the other hand I already have a `getPods` function handy.

pkg/cluster/mpi/statefulset.go
ddl-kevin: What keeps the status updater alive? The pod count? I guess my question is:

Lets say the StatefulSet is able to remove pods (to satisfy the status updater), but maybe not volumes (unlikely, but this is a pathological example). 

This would call delete, and Kubernetes my return "Ok! (but its pending deletion)". Will that be detected anywhere?

I'm not suggesting we block, but requeue if it is only pending and check that it got deleted later.

pkg/cluster/mpi/statefulset.go
ddl-kevin: Yes, I'm referring to the fact that you are now explicitly deleting the StatefulSet. 

Also: Isn't this the purpose of ObjectKeys? Do you know why we are using these partially populated resources to indicate what should be deleted?

pkg/cluster/mpi/statusupdate.go
ddl-kevin: I would say: If the statefulset resource finalizer is done, then the statefulset should not exist, and we can assume all owned resources of the statefulset (which includes pods) do not exist.

pkg/cluster/mpi/statefulset.go
adp312: I need to check this out... In my understanding, the `deleteIfExists` function expects a `StatefulSet` as a "pointer" to what is supposed to be deleted.

adp312: The k8s client does need an `Object` here, not `ObjectKeys`: 
sigs.k8s.io/controller-runtime@v0.10.0/pkg/client/interfaces.go:67

pkg/cluster/mpi/statusupdate.go
adp312: Turns out, we need to check specifically for pods here (not for the stateful set), because the stateful set won't wait for pods if deleted on the foreground.

pkg/cluster/mpi/statefulset.go
adp312: I am going to change this to explicitly tell it to retry.

ddl-kevin (approved):

PR #75, opened 2022-01-19
DOM-35204: MPI Istio Support

ddl-kevin:


pkg/cluster/mpi/service.go
```diff
@@ -9,36 +10,65 @@ import (
...
+		ports = append(ports, corev1.ServicePort{
+			Name:       sshdPortName,
+			Port:       sshdPort,
+			TargetPort: intstr.FromString(sshdPortName),
+			Protocol:   corev1.ProtocolTCP,
+		})
+
+		selector = meta.MatchLabelsWithComponent(cr, c.comp)
+		extraLabels = cr.Spec.Worker.Labels
+	case metadata.ComponentNone:
```
adp312, line 55: `default`?

pkg/cluster/mpi/istiopeerauthentication.go
```diff
@@ -0,0 +1,40 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	dcv1alpha1 "github.com/dominodatalab/distributed-compute-operator/api/v1alpha1"
+	"github.com/dominodatalab/distributed-compute-operator/pkg/cluster/metadata"
+	"github.com/dominodatalab/distributed-compute-operator/pkg/controller/components"
+	"github.com/dominodatalab/distributed-compute-operator/pkg/controller/core"
+	"github.com/dominodatalab/distributed-compute-operator/pkg/resources/istio"
+)
+
+func IstioPeerAuthentication(enabled bool) core.Component {
```
adp312, line 13: Should it return `OwnedComponent`?

adp312 (approved):
Looks good except for the default case in the switch.

pkg/cluster/mpi/service.go
ddl-kevin: I don't really know what default would mean here, because we only create services for specific components so this is undefined and should never happen. I'm actually not sure how to convince Go that this is not needed.

adp312: I just meant that using `deafult:` here instead of `case metadata.ComponentNone:` would make more sense.

pkg/cluster/mpi/istiopeerauthentication.go
ddl-kevin: This seems to be tricky because the object being returned is imported from the Istio library, and so is missing some methods I think (deepcopy?)

PR #76, opened 2022-01-20
DOM-35411 Handle breaking changes in SparkCluster CRD

adp312:


api/v1alpha1/sparkcluster_types.go
```diff
@@ -93,3 +93,11 @@ type SparkClusterList struct {
 func init() {
 	SchemeBuilder.Register(&SparkCluster{}, &SparkClusterList{})
 }
+
+// IsIncompatibleVersion checks if the provided instance of SparkCluster struct
+// has a version compatible with the current one.
+func (sc *SparkCluster) IsIncompatibleVersion() bool {
+	// Unfortunately we can't rely on e.g. Spec.TypeMeta.APIVersion, because of
+	// past breaking changes for which the meta information hasn't been updated.
+	return sc.Spec.Driver.Selector == nil
```
ddl-kevin, line 102: How did you select this field? I think it can be empty in a valid configuration.

adp312:
The driver selector must have a value; see https://github.com/dominodatalab/distributed-compute-operator/blob/5f112b7b011ff3c6bb9655a1a37a2022c261e98e/api/v1alpha1/sparkcluster_webhook.go#L196).

The old CRD doesn't include `Driver` field, so the structure is initialized to default values, which for `Selector` will be `nil`. I supposed that for a valid CRD, and empty `Selector` would produce an empty map, not `nil`. Just tested this and it turned out this is not true: there are `nil`s in both situations (which inhibits validation). So, it's a problem, but a different one.

ddl-kevin (approved):
Ok, sounds good. Were you able to validate this in a live cluster?

adp312:
I mean, my current approach **is indeed problematic**, because a missing selector in an otherwise up-to-date CRD will suddenly inhibit validation, instead of being caught by it. I am looking into what can be done.

I use a life EKS cluster to test these things as I go.

ddl-kevin (approved):
LGTM

PR #77, opened 2022-01-25
DOM-35210 Running MPI worker nodes with non-privileged user

adp312:


pkg/cluster/mpi/configmap.go
```diff
@@ -14,12 +15,47 @@ import (
...
+SSH_PORT=%[1]d
+DOMINO_UID=%[2]d
+DOMINO_GID=%[3]d
+DOMINO_USER=%[4]s
+DOMINO_GROUP=%[5]s
+AUTHORIZED_KEYS_PATH=%[6]s
+
+if ! id $DOMINO_UID >/dev/null 2>&1; then
+    groupadd -g $DOMINO_GID $DOMINO_GROUP
+    useradd -u $DOMINO_UID -g $DOMINO_GID -mN -s /bin/bash $DOMINO_USER
```
JoshBroomberg, line 32: If useradd or su etc are missing, what will happen? I believe in other failure modes, we end up with a cluster stuck in a pending state with workers in a crash loop backoff. Same here?

pkg/cluster/mpi/configmap.go
```diff
@@ -14,12 +15,47 @@ import (
...
+if [ "$(id -nu $DOMINO_UID)" != "$DOMINO_USER" ]; then
+    echo >&2 "User name mismatch"
+    exit 1
+fi
+
+DOMINO_HOME=$(eval echo "~$DOMINO_USER")
+SSH_DIR="$DOMINO_HOME/.ssh"
+mkdir -p "$SSH_DIR"
+
+ssh-keygen -f "$SSH_DIR/ssh_host_key" -N '' -t rsa
```
JoshBroomberg, line 43: https://blog.trailofbits.com/2019/07/08/fuck-rsa/

I think we should use ecdsa - the same standard we use to generate the user key pair

pkg/cluster/mpi/mpi.go
```diff
@@ -1,33 +1,37 @@
...
+	launchScriptPath   = "/opt/domino/bin/mpi-worker-start.sh"
+	launchScriptMode   = 0544 // octal!
 
-	finalizerRetryPeriod = 1 * time.Second
-)
+	// Default parameters of a user account for executing MPI workload.
+	defaultUserID    = 12574
+	defaultUserName  = "domino"
+	defaultGroupID   = 12574
+	defaultGroupName = "domino"
```
JoshBroomberg, line 28: Do we want to call this 'mpi' and then override it via the CR? Just to maintain some semblance of this as an open source product?

pkg/cluster/mpi/mpi.go
```diff
@@ -1,33 +1,37 @@
...
+	// SSH port used by MPI and a name of this port within the service
+	sshdPort     = 2222
+	sshdPortName = "sshd"
 
-	hostFileFilename   = "hostfile"
-	sshdConfigFilename = "sshd_config"
+	// Locations of the mounted files and their modes
+	authorizedKeysPath = "/etc/mpi/authorized_keys"
+	authorizedKeysMode = 0444 // octal!
+	launchScriptPath   = "/opt/domino/bin/mpi-worker-start.sh"
```
JoshBroomberg, line 21: see comment below, perhaps something more generic?

pkg/cluster/mpi/mpi.go
```diff
@@ -1,33 +1,37 @@
...
+	defaultUserID    = 12574
+	defaultUserName  = "domino"
+	defaultGroupID   = 12574
+	defaultGroupName = "domino"
+
+	// Configmap key containing the host file
+	hostFileName = "hostfile"
 
-var (
-	sshAuthorizedKeysPath = filepath.Join(configRootMountPath, sshAuthorizedKeysFilename)
```
JoshBroomberg, line 30: Why did this get deleted?

JoshBroomberg: moved to LOC 19?

pkg/cluster/mpi/statefulset.go
```diff
@@ -26,16 +27,18 @@ var (
...
 	}
 	workerCommand = []string{
-		"/usr/sbin/sshd",
-		"-De",
+		launchScriptPath,
 	}
 )
 
+// Key of the shared Secret object that contains client-side SSH public key
+const publicKeyField = "ssh-publickey"
```
JoshBroomberg, line 40: I am suprirsed this isn't a constant somewhere in the k8s lib, I don't think we set it manually

JoshBroomberg (approved):
LGTM, but wait on approval from Kevin as his head is more in this code than mine. I take it all fields are optional such that we could, in theory, run this without changes to nucleus?

pkg/cluster/mpi/configmap.go
adp312: Correct. The script will fail right away, and Kubernetes will try to recreate pods. There is an outstanding bug to handle this in the controller properly.

pkg/cluster/mpi/configmap.go
adp312: Just checked that `ssh-keygen` supports ecdsa, so fine with me.

pkg/cluster/mpi/mpi.go
adp312: I wasn't sure what the right name should be. The main definition is this: https://github.com/dominodatalab/distributed-compute-operator/blob/77c4378769ac3ace4b078054c41399f42212258b/api/v1alpha1/mpicluster_types.go#L13. The names of these constants should derive from that.

pkg/cluster/mpi/mpi.go
adp312: It wasn't immediately clear in the old design what these paths were constructed from (a bunch of similarly named strings). I reversed this so that "fundamental" value now is the full path, and when it's needed I extract a base name from it (see calls to `filepath.Base(..)` elsewhere).

adp312: Basically, I reduced the number of constants.

pkg/cluster/mpi/statefulset.go
adp312: Agree. There may be.

pkg/cluster/mpi/mpi.go
adp312: Let me know if you have suggestions for better names (for CR fields). Now is a good time to change this.

pkg/cluster/mpi/statefulset.go
ddl-kevin: `corev1.SSHAuthPrivateKey`

pkg/cluster/mpi/configmap.go
```diff
@@ -14,12 +15,47 @@ import (
...
+DOMINO_USER=%[4]s
+DOMINO_GROUP=%[5]s
+AUTHORIZED_KEYS_PATH=%[6]s
+
+if ! id $DOMINO_UID >/dev/null 2>&1; then
+    groupadd -g $DOMINO_GID $DOMINO_GROUP
+    useradd -u $DOMINO_UID -g $DOMINO_GID -mN -s /bin/bash $DOMINO_USER
+fi
+if [ "$(id -nu $DOMINO_UID)" != "$DOMINO_USER" ]; then
+    echo >&2 "User name mismatch"
```
ddl-kevin, line 35: Can we rename the user instead of failing?

pkg/cluster/mpi/configmap.go
```diff
@@ -14,12 +15,47 @@ import (
...
+fi
+
+DOMINO_HOME=$(eval echo "~$DOMINO_USER")
+SSH_DIR="$DOMINO_HOME/.ssh"
+mkdir -p "$SSH_DIR"
+
+ssh-keygen -f "$SSH_DIR/ssh_host_key" -N '' -t rsa
+chmod 400 "$SSH_DIR/ssh_host_key"
+
+cat << EOF > "$SSH_DIR/sshd_config"
```
ddl-kevin, line 46: Is there a reason we moved from mounting this in a configmap? It seems like this could still be mounted statically and this script simply configures the user and then launches `sshd` with `su`. I don't think we actually need the user's home directory, we can just place this wherever and pass that to `sshd` on launch.

ddl-kevin (approved):
Looks good, just a few questions.

pkg/cluster/mpi/configmap.go
adp312: I've never thought renaming a user is an option in Linux, but looks like `usermod` command can do exactly that... 

Still, I think this is a wrong path: as we're requiring UID and username as CRD arguments (with well-known defaults), a mismatch of those is clearly a violation of the requirement and should cause an exception. Unless of course we find some common scenario of when this happens, in which case we may want to silently rename the user.

pkg/cluster/mpi/configmap.go
adp312: Yeah, we can... I just thought it's simpler this way: everything is one place, single mount etc. For the price of an ugly `cat << EOF` construct, which doesn't look too terrible here.

pkg/cluster/mpi/configmap.go
ddl-kevin: Hmmm, what requirements are being violated here? The UID and GID we use is arbitrary. If some image happens to already have that user configured, we can probably use it, but we need a specific username to login, which is largely cosmetic.

pkg/cluster/mpi/configmap.go
ddl-kevin: I guess to me the harness script is a bit of an anti pattern and we should keep it as small as possible. If we don't need any information from the live container to generate the config file, I think is should be mounted read only from a configmap

PR #81, opened 2022-01-28
DOM-35816 Rsync sidecar

adp312:


adp312:
The code is functional, but needs to be refactored a bit.

pkg/cluster/mpi/statefulset.go
```diff
@@ -34,6 +34,26 @@ var (
...
+			TCPSocket: &corev1.TCPSocketAction{
+				Port: intstr.FromInt(rsyncPort),
+			},
+		},
+	}
+	sidecarCommand = []string{
+		"/bin/bash",
+		"-c",
+		"/usr/bin/ssh-keygen -f /opt/domino/etc/ssh/ssh_host_key -N '' -t ecdsa && " +
+			"/usr/sbin/sshd -f /opt/domino/etc/ssh/sshd_config -o \"Port 2223\" -De",
```
adp312, line 54: ^^ TODO: port from a constant

pkg/cluster/mpi/statefulset.go
```diff
@@ -34,6 +34,26 @@ var (
...
+	}
+	sidecarImage = "quay.io/ddl_apetrov/rsync-sidecar:latest"
```
adp312, line 56: ^^^ To be moved to a more appropriate location

pkg/cluster/mpi/mpi.go
```diff
@@ -11,9 +11,9 @@ import (
 
 const (
 
-	// SSH port used by MPI and a name of this port within the service
+	// SSH port used by MPI worker
 	sshdPort     = 2222
-	sshdPortName = "sshd"
+	sshdPortName = "sshd" // service port name
```
ddl-kevin, line 16: Can you remove this comment? It will cause a merge conflict with my PR (the service name is not Istio compatible).

pkg/cluster/mpi/mpi.go
```diff
@@ -27,6 +27,15 @@ const (
 	defaultGroupID   = 12574
 	defaultGroupName = "domino"
 
+	// SSH ports used by rsync sidecar
+	rsyncPort     = 2223
+	rsyncPortName = "rsync" // service port name
```
ddl-kevin, line 32: Can you change this to `tcp-rsync` for Istio compatibility. 

https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

scripts/rsync.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:20.04
```
ddl-kevin, line 1: How large is this image? Can we start from something smaller like `alpine` or `debian` slim? 

Also, I think this should go in a `dockerfiles` folder or something like that.

scripts/rsync.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ARG DOMINO_UID=12574
```
ddl-kevin, line 3: These must be set dynamically based on what is specified in the CRD, right?

scripts/rsync.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ARG DOMINO_DIR=/opt/domino
+ARG DOMINO_SSH_DIR=${DOMINO_DIR}/etc/ssh
+
+ARG SSHD_CONFIG=${DOMINO_SSH_DIR}/sshd_config
+ARG AUTHORIZED_KEYS_PATH=/etc/mpi/authorized_keys
+
+RUN \
+	apt-get update && \
+	apt-get -y install \
```
ddl-kevin, line 18: Make sure to clear the `apt` cache to make the layer as small as possible. 

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

pkg/cluster/mpi/statefulset.go
ddl-kevin: Some notes for productionizing this (cc @JoshBroomberg):

* We should add a GitHub action to publish the images using the same scheme as DCO itself. 
* We will need to figure out how to configure the image the DCO uses. I believe we will need to propagate this config from Helm -> DCO CLI. 
* We need to add this image to the catalog of images in platform apps which is used for pulling images into internal registries during offline installs. 

@adp312 Can you generate some follow-up tickets for this?

ddl-kevin (commented):
Looks like a good start, added some notes.

scripts/rsync.Dockerfile
adp312: Oh, right. And possibly some other cleanup.

scripts/rsync.Dockerfile
adp312: The other way around, I suppose... Here, the IDs are "baked into" the image, so DCO controller must use the same when launching the sidecar. Those values are hardcoded as constants in `pkg/cluster/mpi/mpi.go`. So, there are two places in the code containing hardcoded sets of values that must be consistent with each other. Which is OK.

ddl-kevin: I think this is wrong, because Nucleus still needs to know how to SSH into the `rsync` container. I supposed it can rely on the hardcoding, but that seems brittle.

adp312: Yeah, I see your point. We'd have to create the user in the runtime, similar to the main container, instead of baking it in. But if we must be able to specify the user account via CRD, that's OK. I've just though that my approach will be more elegant (through less flexible).

ddl-kevin: I guess in the interest of avoiding running as root, lets keep this hardcoded in both Nucleus and here (i.e. what you have now).

scripts/rsync.Dockerfile
adp312: 119M with debian:11-slim. Used to be 203-236M with ubuntu.

ddl-kevin (approved):

PR #82, opened 2022-01-31
DOM-35529 MPI cluster Failed status

adp312:


ddl-kevin (dismissed):

pkg/cluster/mpi/statusupdate.go
```diff
@@ -23,8 +23,15 @@ const (
 	StartingStatus v1.JobConditionType = "Starting"
 	RunningStatus  v1.JobConditionType = "Running"
 	StoppingStatus v1.JobConditionType = "Stopping"
+	FailedStatus   v1.JobConditionType = "Failed"
 )
 
+// failureReasons lists generic Pod failure reasons that cause FailedStatus of MPI Cluster,
+// and map them to more specific strings that are reported in MPI CRD.
+var failureReasons = map[string]string{
+	"CrashLoopBackOff": "EntryPointFailed",
```
ddl-kevin, line 32: I'm not sure what all could lead to this condition, can we check like the exit code on the container, or something more specific?

pkg/cluster/mpi/statusupdate.go
```diff
@@ -130,11 +149,27 @@ func getPods(ctx *core.Context, cr *dcv1alpha1.MPICluster) ([]corev1.Pod, error)
...
 	for _, cond := range pod.Status.Conditions {
 		if cond.Type == corev1.PodReady {
-			return cond.Status == corev1.ConditionTrue
+			if cond.Status == corev1.ConditionTrue {
+				return true, ""
+			}
+			break
+		}
+	}
+	for _, containerStatus := range pod.Status.ContainerStatuses {
```
ddl-kevin, line 163: Should we look at specific containers by name? 

There may be sidecars that we aren't thinking about like Istio proxy.

pkg/cluster/mpi/statusupdate.go
```diff
@@ -51,10 +58,17 @@ func (c statusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result, error)
...
 	var runningPodCnt = 0
+	var failureReason = ""
 	for _, pod := range pods {
 		podNames = append(podNames, pod.Name)
-		if isPodReady(pod) {
+		running, podReason := isPodReady(pod)
+		if running {
 			runningPodCnt++
+		} else if failureReason == "" {
+			mpiReason, failure := failureReasons[podReason]
```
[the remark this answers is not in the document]
adp312: Yes, the second one is a boolean telling if the key is present.

pkg/cluster/mpi/statusupdate.go
adp312: Yes, but I intentionally handle a sidecar failure and a main container failure in the same way. I don't think we should distinguish them, as a non-functional rsync is equally bad.

pkg/cluster/mpi/statusupdate.go
adp312: It's possible to obtain container's last exit code, but it doesn't look more informative than `CrashLoopBackOff`. The only advantage I see is that the we know what an exit code is, while exact meanings of Kubernetes events are not well documented. I have another idea, though: In the current design we do not handle legitimate container restarts. That is, MPI's Failed status must only be used if the container has never managed to run. I will work on that.

adp312:
Only changes to `pkg/cluster/mpi/statusupdate.go` are relevant for this work (I accidentally merge DOM-35816 into this branch).

ddl-kevin (approved):

PR #83, opened 2022-02-02
DOM-35163 Validation of MPI Cluster object lifecycle

adp312:


controllers/mpicluster_controller.go
```diff
@@ -16,6 +17,7 @@ import (
 func MPICluster(mgr ctrl.Manager, webhooksEnabled, istioEnabled bool) error {
 	reconciler := core.NewReconciler(mgr).
 		For(&dcv1alpha1.MPICluster{}).
+		Component("istio-peerauthentication", mpi.IstioPeerAuthentication(istioEnabled)).
```
ddl-kevin, line 20: Sadly this will need to be updated from your other refactor. Sorry for the delay in reviewing.

ddl-kevin (approved):

controllers/mpicluster_controller.go
adp312: Yeah, it'll need a slight edit to be merged.

PR #87, opened 2022-02-10
DOM-35895 DCO Deployment

adp312:


deploy/helm/distributed-compute-operator/values.yaml
```diff
@@ -111,5 +111,5 @@ nameOverride: ""
 fullnameOverride: ""
 
 mpi:
-  initImage: "quay.io/domino/distributed-compute-operator-mpi-init:0.0.0"
-  syncImage: "quay.io/domino/distributed-compute-operator-mpi-sync:0.0.0"
+  initImage: quay.io/domino/distributed-compute-operator-mpi-init:main
+  syncImage: quay.io/domino/distributed-compute-operator-mpi-sync:main
```
JoshBroomberg, line 115: WHy don't these use the same split-up format as the one above?

JoshBroomberg (approved):
One nit, but looks good.

deploy/helm/distributed-compute-operator/templates/deployment.yaml
```diff
@@ -69,10 +69,10 @@ spec:
             - --istio-enabled
             {{- end }}
             {{- with .Values.mpi.initImage }}
-            - --mpi-init-image={{ . }}
+            - --mpi-init-image="{{ .registry }}/{{ .repository }}:{{ .tag }}"
```
JoshBroomberg, line 72: I believe you can do `{{ include "common.imagename" .Values.mpi.initImage }}`

JoshBroomberg: This is the point of using the split up OCI format

deploy/helm/distributed-compute-operator/values.yaml
```diff
@@ -57,11 +57,11 @@ prometheus:
   namespaceLabels: {}
 
 image:
-  registry: ghcr.io
+  registry: quay.io
   repository: dominodatalab/distributed-compute-operator
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+  tag: main
```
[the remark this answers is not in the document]
JoshBroomberg: Main. No tag will use latest - which is not what we want

JoshBroomberg (commented):
LIttle nit. Otherwise I think we should merge this and then work out a solution for the image sync problem (it is solved in an ok way for now)

deploy/helm/distributed-compute-operator/templates/deployment.yaml
adp312: should it be "common.images.image"? https://artifacthub.io/packages/helm/bitnami/common#images

adp312:
Tested everything. Will commit next Monday.

PR #92, opened 2022-02-17
DOM-36224 Mount Kerberos keytab on MPI workers

adp312:


.gitignore
```diff
@@ -27,3 +27,6 @@ testbin/*
 # goreleaser directories
 dist/
 custom-resource-definitions/
+
+# miscellaneous local files
+exclude/
```
adp312, line 32: Unrelated to this work

pkg/cluster/mpi/configmap.go
```diff
@@ -33,6 +33,11 @@ func (c configMapComponent) Reconcile(ctx *core.Context) (ctrl.Result, error) {
 			hostFileName: buildHostFile(cr),
```
ddl-kevin, line 33: Hmmm, is this still used? Or was the `ConfigMap` unused before.

pkg/cluster/mpi/statefulset.go
```diff
@@ -209,14 +212,32 @@ func secretVolumes(cr *dcv1alpha1.MPICluster) ([]corev1.Volume, []corev1.VolumeM
 				},
 			},
 		},
+		{
```
ddl-kevin, line 215: Would it be possible to only create the `ConfigMap`, `Volume`, `VolumeMount` if the keytab field in the CR is populated?

ddl-kevin (changes requested):

pkg/cluster/mpi/configmap.go
adp312: Not sure if it's actually used, but I though it's a good idea to keep an "authoritative" hostfile generated on the server in case it's needed, because it's an essential piece of data for MPI. There is no harm to have it in any case. (If I am not mistaken, in Nucleus we ended up creating the source file be deducing names of the nodes that will be in the cluster.)

adp312: (1) One map per "area of concerns"?
(2) I think the way it works now is bad. This should be a secret.

ddl-kevin: > (1) One map per "area of concerns"?

Seems reasonable to me (1 is mounted in a client (or launcher before), 1 in a worker).

PR #96, opened 2022-03-08
Revert "DOM-36169 Set user home dir"

JoshBroomberg:
Reverts dominodatalab/distributed-compute-operator#95


adp312:
@JoshBroomberg and I agreed that the original problem has been caused by a transient inconsistency of images in the environment. This PR (i.e., the rollback) is likely not needed, but let's keep it around until the situation is confirmed.

JoshBroomberg:
@adp312 confirmed transient. Closing

PR #99, opened 2022-03-10
[DOM-36776] Add repos module to rsyncd.conf

pocheung1:
Add repos module to rsyncd.conf to sync /repos

https://dominodatalab.atlassian.net/browse/DOM-36776

### DIFF
--- a/dockerfiles/rsyncd.conf
+++ b/dockerfiles/rsyncd.conf
@@ -1,9 +1,12 @@
 pid file = $RSYNC_RUN_DIR/rsync.pid
 lock file = $RSYNC_RUN_DIR/rsync.lock
 log file = /dev/stdout
-
-[mnt]
-path = /mnt
 use chroot = false
 read only = false
 timeout = 300
+
+[mnt]
+path = /mnt
+
+[repos]
+path = /repos


dockerfiles/rsyncd.conf
```diff
@@ -1,9 +1,12 @@
 pid file = $RSYNC_RUN_DIR/rsync.pid
 lock file = $RSYNC_RUN_DIR/rsync.lock
 log file = /dev/stdout
-
-[mnt]
-path = /mnt
 use chroot = false
 read only = false
 timeout = 300
```
adp312, line 6: I cannot be completely sure, but at least my reading of the rsyncd.conf man page is that these 3 parameters (`use chroot`, `read only`, `timeout`) are for the modules only.  Also, this is what I saw in examples. Unless we want to go deep in testing of rsyncd, I suggest to copy them into each module individually.

adp312 (changes requested):
Please copy parameters to each module.

dockerfiles/rsyncd.conf
pocheung1: I am able to rsync to /repos with this config file.  I have also tried toggling `use chroot` and `read only` parameters to true and that causes rsync to fail.  So the global parameters are working as intended.

pocheung1: Here is an example here where some module parameters are configured as global:
https://linuxconfig.org/how-to-setup-the-rsync-daemon-on-linux

adp312 (approved):
OK then.

PR #100, opened 2022-03-11
Enable descriptions in CRD manifests

adp312:
No functional changes here.


ddl-kevin (approved):
LGTM. Any idea why Sonny disabled this?

adp312:
I understand that the new version of the manifest generator uses a completely different set of options, and we require now to specifically enable the description field.

PR #101, opened 2022-03-15
[DOM-36776] Add imported module to rsyncd.conf

pocheung1:
Add imported module to rsyncd.conf to sync /mnt/imported

https://dominodatalab.atlassian.net/browse/DOM-36776


JoshBroomberg (approved):
No problem with nested modules in rsync right? (imported is inside mnt)

PR #104, opened 2022-03-17
DOM-37092 add groups from initial user to domino user

JoshBroomberg:
This 'dynamically' adjusts the user we create for the permissions on k8s-mounted file systems. It does this by adding our user to all of the non-root groups to which the container-user (root in our current setup) belongs

### DIFF
--- a/dockerfiles/mpi-worker-start.sh
+++ b/dockerfiles/mpi-worker-start.sh
@@ -37,8 +37,9 @@ fi
 for gid in `id -G`; do
   if [ $gid != 0 ]; then
     # Add user to a new/existing group with desired id.
+    # https://askubuntu.com/a/639998
     group_name=$(cut -d: -f1 < <(getent group $gid))
-    if [ -z $group_name ]; then
+    if [ -z "$group_name" ]; then
         group_name="group-$gid"
 	    groupadd -g $gid $group_name
     fi
--- a/dockerfiles/mpi-worker-start.sh
+++ b/dockerfiles/mpi-worker-start.sh
@@ -20,13 +20,29 @@ fi
 if ! id $DOMINO_UID >/dev/null 2>&1; then
 	useradd -u $DOMINO_UID -g $DOMINO_GID -mN -s /bin/bash -d "$DOMINO_HOME_DIR" $DOMINO_USER
 else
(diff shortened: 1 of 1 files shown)


dockerfiles/mpi-worker-start.sh
```diff
@@ -20,14 +20,30 @@ fi
...
+    usermod -a -G $DOMINO_GROUP $DOMINO_USER
 fi
 
+# Add the new domino user to the non-root groups of the current container user
+for gid in `id -G`; do
+  if [ $gid != 0 ]; then
+    gname="group-$gid"
+    if ! cut -d: -f3 < /etc/group | grep "^${gid}$" >/dev/null 2>&1; then
+	    groupadd -g $gid $gname
+    fi
```
adp312, line 42: ... else if $gid is found in the password file, we should get a $gname associated with it.

adp312 (changes requested):
Unless I miss something, the logic is not complete

dockerfiles/mpi-worker-start.sh
JoshBroomberg: Thank you.

dockerfiles/mpi-worker-start.sh
```diff
@@ -20,14 +20,32 @@ fi
...
+
+    # Add to domino group (idempotent)
+    usermod -a -G $DOMINO_GROUP $DOMINO_USER
 fi
 
+# Add the new domino user to the non-root groups of the current container user
+for gid in `id -G`; do
+  if [ $gid != 0 ]; then
+    # Add user to a new/existing group with desired id.
+    group_name=$(cut -d: -f1 < <(getent group $gid))
```
adp312, line 40: I am not exactly sure what `<(` does, but it seems to work.

dockerfiles/mpi-worker-start.sh
```diff
@@ -20,14 +20,32 @@ fi
...
+    if [ -z $group_name ]; then
```
adp312, line 41: Quotes! `[ -z "$group_name" ]`

adp312 (approved):
Add quotes. Otherwise OK.

PR #105, opened 2022-03-17
DOM-37056 Configure bash environment for Domino user

adp312:


JoshBroomberg (approved):

adp312:
A reason of the troubles: https://dominodatalab.atlassian.net/browse/DOM-37056?focusedCommentId=273125

adp312:
Successfully tested with Horovod and Tensorflow here: https://mpiist12042.gambit-sandbox.domino.tech/workspace/integration-test/quick-start

dockerfiles/mpi-worker-start.sh
```diff
@@ -47,6 +48,20 @@ for gid in `id -G`; do
...
+cat << EOF > "$DOMINO_HOME_DIR/.profile"
+if [ "\$BASH" ] && [ -f ~/.bashrc ]; then
+  . ~/.bashrc
+fi
+EOF
+chmod 644 "$DOMINO_HOME_DIR/.profile"
+
+rm -f "$DOMINO_HOME_DIR/.bashrc"
+touch "$DOMINO_HOME_DIR/.bashrc"
+printenv | grep PATH | sed 's;^;export ;' >> "$DOMINO_HOME_DIR/.bashrc"
```
ddl-kevin, line 60: I take it that it is a conscious decision to only copy certain vars? We are intentionally not copying `ENV` vars from the image? Or do those already work?

dockerfiles/mpi-worker-start.sh
```diff
@@ -47,6 +48,20 @@ for gid in `id -G`; do
...
+printenv | grep MPI | sed 's;^;export ;' >> "$DOMINO_HOME_DIR/.bashrc"
+printenv | grep DOMINO | sed 's;^;export ;' >> "$DOMINO_HOME_DIR/.bashrc"
+chmod 644 "$DOMINO_HOME_DIR/.bashrc"
```
ddl-kevin, line 63: This is read-only by the `domino` user right?

ddl-kevin (approved):
LGTM. Just some questions.

dockerfiles/mpi-worker-start.sh
adp312: Yes, we copying only relevant variables. The source of truth is the current **runtime environment** (which is ENV + whatever bash init files got sourced).

dockerfiles/mpi-worker-start.sh
adp312: Oops. The `.profile` and `.bashrc` should be owned by domino. Will fix now. A mode probably doesn't matter, and I set to to 644 only because it's a usual mode for this kind of files.

PR #106, opened 2022-03-18
Create PeerAuthentication for cluster client when Istio disabled

ddl-kevin:
When Istio injection is disabled in the cluster in an Istio-enabled deployment, the workers cannot communicate back to the client which still has an Istio sidecar expecting mTLS. 

We detect this and create a PERMISSIVE PeerAuthentication for the _client_ to allow this communication.

### DIFF
--- a/pkg/cluster/mpi/istiopeerauthentication.go
+++ b/pkg/cluster/mpi/istiopeerauthentication.go
@@ -8,6 +8,7 @@ import (
 	"github.com/dominodatalab/distributed-compute-operator/pkg/controller/components"
 	"github.com/dominodatalab/distributed-compute-operator/pkg/controller/core"
 	"github.com/dominodatalab/distributed-compute-operator/pkg/resources/istio"
+	authenticationv1alpha1 "istio.io/api/authentication/v1alpha1"
 )
 
 func IstioPeerAuthentication(enabled bool) core.Component {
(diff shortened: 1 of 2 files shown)


pkg/cluster/mpi/istiopeerauthentication.go
```diff
@@ -38,3 +39,32 @@ func (s *istioPeerAuthenticationDS) Enabled() bool {
...
+	return &istio.PeerAuthInfo{
+		Name:      meta.InstanceName(s.mpi, ComponentClient),
+		Namespace: s.mpi.Namespace,
+		Labels:    meta.StandardLabels(s.mpi),
+		Selector:  s.mpi.Spec.NetworkPolicy.ClientLabels,
+		Mode:      authenticationv1alpha1.MutualTls_PERMISSIVE.String(),
+	}
+}
+
+func (s *istioClientPeerAuthenticationDS) Enabled() bool {
```
JoshBroomberg, line 64: nit: possible to use the internal bool /  setting that results in this annotations value rather than the annotation itself? Would be more resilient to changes in API in the future

ddl-kevin: That is not currently passed in the CR. Principally, we are more worried about whether the sidecar is being injected than how that is configured. Unfortunately the annotation is probably the most stable way to determine that.

adp312 (approved):

PR #107, opened 2022-03-18
Restore pull policy to 'if not present'

JoshBroomberg:
Undoing our floating-main development approach


adp312 (approved):
As of your question in JIRA -- you're correct: we ended up having "Always" only in this place. Nothing else needs to be reverted, AFAIK.

PR #109, opened 2022-05-18
DOM-37301: Enable addition of worker container security context capabilities

ddl-kevin:


pkg/cluster/dask/statefulset.go
```diff
@@ -147,7 +159,7 @@ func (s *statefulSetDS) imagePullSecrets() []corev1.LocalObjectReference {
 	return s.dc.Spec.ImagePullSecrets
 }
 
-func (s *statefulSetDS) securityContext() *corev1.PodSecurityContext {
+func (s *statefulSetDS) podSecurityContext() *corev1.PodSecurityContext {
```
JoshBroomberg, line 162: This alone is enough reason for me to keep being OCD about names

JoshBroomberg (approved):

PR #110, opened 2022-05-25
RM-1305 Update debian OS libs in MPI images

nakkurt:
### What problem does this PR solve?
* We need to update the base Debian image tag for `*-mpi-*` images to bring in OS lib updates with security fixes from upstream.
    * these images need to be updated as part of the [release OSRP Process](https://dominodatalab.atlassian.net/wiki/spaces/ENG/pages/1938522147/OSRP+Process) going forward and we need this for the currently ongoing 5.1.3 release.
    * More context: [RM-1305](https://dominodatalab.atlassian.net/browse/RM-1305)

### What is the solution?
* Update the base Debian image tag to `quay.io/domino/debian:10.11-20220520-1846`, which includes updated OS libs from upstream Debian's package manager.

### DIFF
--- a/dockerfiles/mpi-init.Dockerfile
+++ b/dockerfiles/mpi-init.Dockerfile
@@ -60,7 +60,7 @@ RUN \
 
 # The final image only contains built artifacts.
 # The base image should be up-to-date, but a specific version is not important.
-FROM quay.io/domino/debian:10.11-20220210-1721
+FROM quay.io/domino/debian:10.11-20220520-1846
 WORKDIR /root
 COPY --from=0 /root/worker-utils.tgz ./
 CMD tar -C / -xf /root/worker-utils.tgz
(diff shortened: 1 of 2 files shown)


ddl-kevin:
Imo we can just merge to main. The change I introduced added an optional field so there should be no issues including it in 5.1.3. 

Once you merge, you can simply create a tag of 0.6.2, which will publish images and charts. 

You can then update the catalog to use that new version for everything after 5.1.3.

In theory we could only update the MPI images in the catalog if you are worried about the code changes to the DCO.

I will leave it to @adp312 to approve since he built these images.

nakkurt:
> In theory we could only update the MPI images in the catalog if you are worried about the code changes to the DCO.

Yep, my plan is to use this PR for updating only the **two MPI images** in the 5.1.3 Catalog

adp312 (approved):
All good

PR #113, opened 2022-07-08
bump helm to 3.9.3 to publish in the new OCI format

steved:
Helm 3.7 changed the default chart format when publishing to an OCI repo. This is backwards-incompatible with Helm < 3.7, so we've been publishing using 3.6 for now until we can transition fully to Helm >= 3.7.


sonnysideup (approved):
LGTM. This is essentially the same thing I'm doing in https://github.com/dominodatalab/hephaestus/blob/f3de88dbd66a0a7da4df632c31af73b24538580c/.github/workflows/main.yml#L180-L210, and your charts are pushing successfully, so #profit.

PR #115, opened 2022-08-08
add the envoyfilter template and defaults that allows setting istio h…

ddl-ryan-connor:
**Problem**
Some workloads that use `distributed-compute-operator` (DCO) have an HTTP communication pattern that leaves HTTP connections open but unused for longer than an hour. But, Istio has a default HTTP idle timeout of one hour, after which it closes such connections, which can cause unrecoverable errors in the workload. One example is using PyTorch with Ray: Each worker opens an HTTP connection to the head at the beginning of training, and then only uses that connection at the start of each training epoch. So, if Istio is enabled and if an epoch lasts longer than an hour, Istio will close the workers' connections, which can prevent completion of the training. See [this Jira ticket](https://dominodatalab.atlassian.net/browse/DOM-36271) for more info.

**Solution**
This PR introduces the `istio.httpIdleTimeout.timeout` variable to the DCO chart. If Istio is enabled and the user provides a value for the Istio HTTP timeout, then an Istio `EnvoyFilter`(s) is created in the appropriate namespace(s) that will set the HTTP timeout of pods managed by the DCO.

**How was the solution tested?**
- I manually ran `kubectl apply -f file_that_specifies_the_envoy_filter.yaml` on an existing k8s cluster and confirmed that the HTTP timeout was correctly set on new DCO-managed pods by inspecting the Istio proxy pod config dump. I also confirmed the communication error mentioned on the "problem" section went away on a new PyTorch training workload.
- I manually tested constructing the DCO chart with various combinations of values to ensure it built the expected resources using `helm install mytest ./deploy/helm/distributed-compute-operator --dry-run --debug -f myvals.yaml`.

**References**
[Istio docs for EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/)
[Background/starting point for this solution](https://stackoverflow.com/questions/63843610/istio-proxy-closing-long-running-tcp-connection-after-1-hour)

### DIFF
--- a/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
+++ b/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
@@ -3,11 +3,10 @@
 {{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
 {{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
 {{- range $namespace := $namespaces }}
----
 apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
-  name: distributed-compute-cluster-envoyfilter
+  name: {{ include "common.names.fullname" . }}-envoyfilter
   namespace: {{ $namespace }}
 spec:
   workloadSelector:
@@ -41,6 +40,5 @@ spec:
         typed_config:
           '@type': type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy
           idle_timeout: {{ $.Values.istio.httpIdleTimeout.timeout }}
----
 {{- end }}
 {{- end }}
--- a/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
+++ b/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
@@ -1,0 +1,46 @@
+{{- if and (.Values.istio.enabled) (.Values.istio.httpIdleTimeout.timeout)  }}
+{{- $justRootConfigNamespace := list .Values.istio.rootConfigNamespace }}
+{{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
+{{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
+{{- range $namespace := $namespaces }}
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: distributed-compute-cluster-envoyfilter
+  namespace: {{ $namespace }}
+spec:
+  workloadSelector:
+    labels:
+      app.kubernetes.io/managed-by: distributed-compute-operator
+  configPatches:
+  - applyTo: NETWORK_FILTER
+    match:
+      listener:
+        filterChain:
+          filter:
+            name: "envoy.filters.network.http_connection_manager"
+    patch:
+      operation: MERGE
+      value:
+        name: "envoy.filters.network.http_connection_manager"
+        typed_config:
+          "@type": "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager"
+          common_http_protocol_options:
+            idle_timeout: {{ $.Values.istio.httpIdleTimeout.timeout }}
+  - applyTo: NETWORK_FILTER
+    match:
+      listener:
+        filterChain:
+          filter:
+            name: envoy.filters.network.tcp_proxy
+    patch:
+      operation: MERGE
+      value:
+        name: envoy.filters.network.tcp_proxy
+        typed_config:
+          '@type': type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy
+          idle_timeout: {{ $.Values.istio.httpIdleTimeout.timeout }}
+---
+{{- end }}
+{{- end }}
--- a/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
+++ b/deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
@@ -1,4 +1,5 @@
 {{- if and (.Values.istio.enabled) (.Values.istio.httpIdleTimeout.timeout)  }}
+{{- $commonFullname := include "common.names.fullname" . }}
 {{- $justRootConfigNamespace := list .Values.istio.rootConfigNamespace }}
 {{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
 {{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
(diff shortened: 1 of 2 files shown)


deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if and (.Values.istio.enabled) (.Values.istio.httpIdleTimeout.timeout)  }}
+{{- $justRootConfigNamespace := list .Values.istio.rootConfigNamespace }}
+{{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
+{{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
+{{- range $namespace := $namespaces }}
+---
```
sonnysideup, line 6:
```suggestion
```

✂️

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
```diff
@@ -0,0 +1,46 @@
...
+          filter:
+            name: envoy.filters.network.tcp_proxy
+    patch:
+      operation: MERGE
+      value:
+        name: envoy.filters.network.tcp_proxy
+        typed_config:
+          '@type': type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy
+          idle_timeout: {{ $.Values.istio.httpIdleTimeout.timeout }}
+---
```
sonnysideup, line 44:
```suggestion
```

✂️

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if and (.Values.istio.enabled) (.Values.istio.httpIdleTimeout.timeout)  }}
+{{- $justRootConfigNamespace := list .Values.istio.rootConfigNamespace }}
+{{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
+{{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
+{{- range $namespace := $namespaces }}
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: distributed-compute-cluster-envoyfilter
```
sonnysideup, line 10:
```suggestion
  name: {{ include "common.names.fullname" . }}-envoyfilter
```

This name should be unique to prevent collisions if there are multiple releases of DCO. Unlikely, yet possible.

deploy/helm/distributed-compute-operator/values.yaml
```diff
@@ -38,6 +38,18 @@ istio:
...
+  # set the HTTP connection idle timeout for pods managed by the distributed-compute-operator
+  httpIdleTimeout:
+    # the timeout to set. Leave empty for istio default.
+    timeout: ""
+    # If timeout and namespaces are set, sets the timeout only in given namespaces.
+    # If timeout is set but namespaces is empty:
+    #   - if config.watchNamespaces is empty, then timeout is set across the whole istio mesh
+    #   - else timeout is set for namespaces in config.watchNamespaces
+    # If any namespaces in the istio mesh are unaffected by this setting, they get istio default.
+    namespaces: []
```
sonnysideup, line 52: I don't think setting a timeout _across the whole istio mesh_ belongs here (i.e. this type of config belongs in the installer code). Creating a localized, targeted config is fine but not a cluster change.

I'm assuming `watchNamespaces` and `namespaces` are adequate, right?

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if and (.Values.istio.enabled) (.Values.istio.httpIdleTimeout.timeout)  }}
+{{- $justRootConfigNamespace := list .Values.istio.rootConfigNamespace }}
+{{- $defaultNamespaces := .Values.config.watchNamespaces | default $justRootConfigNamespace }}
+{{- $namespaces := .Values.istio.httpIdleTimeout.namespaces | default $defaultNamespaces }}
+{{- range $namespace := $namespaces }}
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
```
sonnysideup, line 8: qq: why create this with the helm chart and not add the resource creation to the code? just curious.

deploy/helm/distributed-compute-operator/values.yaml
ddl-ryan-connor: since the default `watchedNamespaces` is empty list which means "all namespaces", then the appropriate default for this http idle timeout is "across the whole istio mesh for pods managed by the DCO". that's what i was hoping the comment on line 43 communicated -- that its only for pods managed by the DCO

ddl-ryan-connor: what if i added "for pods managed by the DCO" to the end of that line 49 comment?

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: if you were going to add the resource creation to the code instead, where and how would you do that? what are the pros and cons of that?

asking because i dont have a great answer for this

ddl-ryan-connor: (i did consider it but didn't find a good place in the code to do this -- i might be missing something)

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: done

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: done

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: done

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: also given the nature of the issue it seemed like a good setting to have as part of the base chart

sonnysideup: Well, you'd have to change the resources for a given cluster type (e.g. `RayCluster`) if it's only limited to Ray. If it's required by everything, then this approach is the easier one. The only limiting aspect of this would be if you wanted to set different timeout values for say ray, dask, spark, etc... In that case, the code approach would be better IMO.

sonnysideup (approved):

deploy/helm/distributed-compute-operator/templates/http-timeout-envoyfilter.yaml
ddl-ryan-connor: ok. right now i think its best to apply to all the cluster types, dont really have a use case for different vals per cluster type. good to keep in mind for the future

PR #121, opened 2022-09-12
(5.3) Check if replicas value is nil before setting config

noahjax:
### Description
We used to have 
```
if csc.WorkerReplicas != *sts.Spec.Replicas { // NOTE: panic: runtime error: invalid memory address or nil pointer dereference
	csc.WorkerReplicas = *sts.Spec.Replicas
	modified = true
}
```
If `sts.Spec.Replicas` is `nil`, this code will try to dereference the `nil` pointer. It appears like some changes made in the 5.3 timeline for hybrid may leave the value as `nil` when it is 0, so the fix is just to default the pointer to 0.

### Jira
https://dominodatalab.atlassian.net/browse/DOM-40865

### DIFF
--- a/pkg/controller/components/clusterstatusupdate.go
+++ b/pkg/controller/components/clusterstatusupdate.go
@@ -8,6 +8,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/components/clusterstatusupdate.go
```diff
@@ -72,8 +73,9 @@ func (c *clusterStatusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result
 		csc.WorkerSelector = selector.String()
 		modified = true
 	}
-	if csc.WorkerReplicas != *sts.Spec.Replicas { // NOTE: panic: runtime error: invalid memory address or nil pointer dereference
-		csc.WorkerReplicas = *sts.Spec.Replicas
+	replicas := pointer.Int32Deref(sts.Spec.Replicas, 0)
+	if csc.WorkerReplicas != replicas { // NOTE: panic: runtime error: invalid memory address or nil pointer dereference (see DOM-40865)
+		csc.WorkerReplicas = replicas
```
sonnysideup, line 78:
```suggestion
	if replicas := pointer.Int32Deref(sts.Spec.Replicas, 0); csc.WorkerReplicas != replicas {
		csc.WorkerReplicas = replicas
```

sonnysideup: Any reason not to use a scoped predicate variable here? Also, in my mind, the comment about panics is not terribly relevant at this piont because it's obvious that a nil pointer will panic when accessed, but this no longer has that issue.

sonnysideup (approved):

PR #124, opened 2022-09-20
HBD-575 • Cluster status reporting

adp312:


.golangci.yml
```diff
@@ -34,7 +34,7 @@ linters-settings:
   misspell:
     locale: US
   nestif:
-    min-complexity: 4
+    min-complexity: 6
```
adp312, line 37: With complexity 4, golint complained about pretty innocent code.

api/v1alpha1/common_types.go
```diff
@@ -153,6 +157,10 @@ type WorkloadConfig struct {
 // ClusterStatusConfig defines the observed state of a given cluster. The
 // controllers will generate and populate these fields during reconciliation.
 type ClusterStatusConfig struct {
+	ClusterStatus batchv1.JobConditionType `json:"clusterStatus"`
```
ddl-kevin, line 160: This might need `omitempty` to handle existing clusters?

api/v1alpha1/common_types.go
```diff
@@ -162,3 +170,20 @@ type ClusterStatusConfig struct {
 	// WorkerSelector is the `scale.status.selector` subresource field.
 	WorkerSelector string `json:"workerSelector,omitempty"`
 }
+
+const (
+	PendingStatus  batchv1.JobConditionType = "Pending"
```
ddl-kevin, line 175: I know this was already defined, but this type seems wrong. Its an alias for a `string`, but only some of these enum variants are actually defined. So IDK if there is value in using this type.

controllers/sparkcluster_controller.go
```diff
@@ -447,13 +468,51 @@ func (r *SparkClusterReconciler) reconcileStatefulSets(ctx context.Context, sc *
...
 
+func (r *SparkClusterReconciler) getMasterPod(ctx context.Context, sc *dcv1alpha1.SparkCluster) (corev1.Pod, bool) {
+	opts := []client.ListOption{
+		client.InNamespace(sc.Namespace),
+		client.MatchingLabels(spark.MetadataLabelsWithComponent(sc, spark.ComponentMaster)),
+	}
+	pods := &corev1.PodList{}
+	err := r.List(ctx, pods, opts...)
+	if err != nil && !apierrors.IsNotFound(err) {
+		r.getLogger(ctx).V(1).Error(err, "reading master status")
```
ddl-kevin, line 508: If we are here, will `Items` be defined below? Or should we just return empty list / false?

controllers/sparkcluster_controller.go
```diff
@@ -447,13 +468,51 @@ func (r *SparkClusterReconciler) reconcileStatefulSets(ctx context.Context, sc *
 		log.V(1).Info("updating status", "path", ".status.workerSelector", "value", sc.Status.WorkerSelector)
 	}
 
+	var status v1.JobConditionType
+	masterPod, masterPodFound := r.getMasterPod(ctx, sc)
+	if masterPodFound {
+		if dcv1alpha1.IsPodReady(masterPod) {
```
ddl-kevin, line 474: Does the Pod have the same health check defined on it that Nucleus uses? I wouldn't want to change the when the Pod is considered ready.

api/v1alpha1/common_types.go
adp312: Yeah, I think there should be a separate (our own) custom type based on string.

controllers/sparkcluster_controller.go
adp312: I am pretty sure that `Items` is initialized an en empty slice. I will test and/or correct.

controllers/sparkcluster_controller.go
```diff
@@ -447,13 +468,51 @@ func (r *SparkClusterReconciler) reconcileStatefulSets(ctx context.Context, sc *
...
+		}
+	} else {
+		status = dcv1alpha1.PendingStatus
+	}
+	if sc.Status.ClusterStatus != status && !hasDeletionTimestamp(sc) {
+		updateStatus = true
+		sc.Status.ClusterStatus = status
+		if status == dcv1alpha1.RunningStatus {
+			tt := metav1.Now()
+			sc.Status.StartTime = &tt
```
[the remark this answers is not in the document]
adp312: Correct. The idea was to save when the cluster became usable. If it changes the state from `Running` to something else, the field is reset. The behavior after the cluster is deleted is to match the latter for the sake of completeness.

I believe it's not used for anything right now.

controllers/sparkcluster_controller.go
adp312: Nop, Nucleus appears to use the same logic:

`KubernetesApiHelper.getPodConditions(pod).exists(condition => condition.getType == "Ready" && condition.getStatus == "True")`

pkg/controller/components/clusterstatusupdate.go
```diff
@@ -87,10 +99,48 @@ func (c *clusterStatusUpdateComponent) Reconcile(ctx *core.Context) (ctrl.Result
...
+
 	// only update when fields have changed
 	if modified {
 		err = ctx.Client.Status().Update(ctx, ctx.Object)
 	}
 
 	return ctrl.Result{}, err
 }
+
+func (c clusterStatusUpdateComponent) Finalize(ctx *core.Context) (ctrl.Result, bool, error) {
```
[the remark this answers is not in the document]
adp312: That's what we used for MPI clusters (which is based on the same framework). I don't remember exact details, but it was something like that `Reconcile` has not been called after the deletion timestamp was set.
I found somewhat similar behavior with Ray/Spark (based on the old implementation): after the finalizer is executed, the reconciler couldn't set status anymore.

controllers/raycluster_controller.go
```diff
@@ -304,6 +316,15 @@ func (r *RayClusterReconciler) reconcileAutoscaler(ctx context.Context, rc *dcv1
 	return nil
 }
 
+func (r *RayClusterReconciler) initializeStatus(ctx context.Context, rc *dcv1alpha1.RayCluster) error {
```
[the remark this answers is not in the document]
adp312: I don't think so because `RayCluster` and `SparkCluster` types are technically unrelated. I know I can change this, but it's probably too much work to remove this little duplication (plus, this framework has been abandoned by Sonny anyway).

api/v1alpha1/common_types.go
adp312: Yeah, that would be cleaner. I will change.

On the other hand, Nucleus doesn't complain on the extra field, and CRDs are super-bloated anyway.

ddl-kevin: I'm just thinking this would be where something would break if DCO tries to deserialize an existing cluster and the field is considered "required"

ddl-kevin (approved):

PR #142, opened 2023-04-23
[DOM-46168] Fix MPI script when gid 12574 exists

ddl-ebrown:
[DOM-46168](https://dominodatalab.atlassian.net/browse/DOM-46168)

- Fixes a bug in https://github.com/dominodatalab/distributed-compute-operator/pull/104 / 0458de27e6b86c362dc1cdd17e62608c5dc11b15
   The call to add $DOMINO_USER to $DOMINO_GROUP fails when gid 12574 is
   already in use. Rename the existing group with gid 12574 when it
   exists in the same way the user with id 12574 is renamed.

   All existing Domino MPI images cannot be launched with the DCO as
   they all define user / group ubuntu / ubuntu with uid / gid 12574 /
   12574.

   This fixes launching MPI clusters with the following images:

   ubuntu18-py3.8-r4.1-mpi4.1.2-domino5.1.1
   ubuntu20-py3.9-r4.2-mpi4.1.3-domino5.2
   ubuntu20-py3.9-r4.2-mpi4.1.3-domino5.3
   ubuntu20-py3.9-r4.2-mpi4.1.4-domino5.4
   ubuntu20-py3.9-r4.2-mpi4.1.4-domino5.5
   ubuntu20-py3.9-r4.2-mpi4.1.4-domino5.6

   With those images in use, the init image would previously fail to
   start with the message:

   > usermod: group 'domino' does not exist

   With this fix, the init images run successfully with:

```
   Generating public/private ed25519 key pair.
   Your identification has been saved in /opt/domino/mpi-cluster/etc/ssh_host_key
   Your public key has been saved in /opt/domino/mpi-cluster/etc/ssh_host_key.pub
   The key fingerprint is:
   SHA256:tuRg8Hv9U3zrUvqfLkIem/l5CCMi5aql+1eQLvxCf6w root@mpi-644312881e14a4746ab91c26-mpi-worker-0
   The key's randomart image is:
   +--[ED25519 256]--+
   |                 |
   |                 |
   |    .   .        |
   |     o +         |
   |    . B S    .   |
   |     * X = =  o..|
   |    ..B B = B.+..|
   |    oo = o B.=.o.|
   |   ++.oEo   ++B=o|
   +----[SHA256]-----+
   Server listening on 0.0.0.0 port 2222.
   Server listening on :: port 2222.
   kex_exchange_identification: Connection closed by remote host
   Connection closed by 10.0.37.26 port 58854
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ SHELL := /bin/bash
 
 # Image URL to use all building/pushing image targets
 IMG ?= ghcr.io/dominodatalab/distributed-compute-operator:latest
+MPI_INIT_IMG ?= quay.io/domino/distributed-compute-operator-mpi-init:latest
+MPI_SYNC_IMG ?= quay.io/domino/distributed-compute-operator-mpi-sync:latest
 # Produce CRDs that work with Kubernetes 1.16+ and supports defaulting, api
 # version conversion, and field pruning.
 CRD_OPTIONS ?= "crd:crdVersions=v1,maxDescLen=100"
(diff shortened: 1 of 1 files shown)


noahjax (approved):
Interesting, I never realized that the init and sync images were also necessary to support mpi. Still not 100% sure how they fit in with the other environment images we ship, but overall PR looks good

PR #155, opened 2023-10-16
[DOM-50795] -- remove bad envoyfilter

saahildhulla:


fraenkel (approved):
is there a reason to delete vs fix the envoyfilter?

saahildhulla:
seeing that it was not effective as it stood, I figured there was no point in having it altogether

